PR #4, opened 2020-07-23
Make the slow threshold to log query configurable

inespo:
The threshold to log slow query is static, set at 2.5 seconds.
Unfortunately, different use cases would require different threshold

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,9 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JAVA-IOGRPC-571957:
+    - '*':
+        reason: No replacement available
+        expires: 2020-08-31T00:00:00.000Z
+patch: {}
--- a/gateway-service/src/main/resources/configs/common/application.conf
+++ b/gateway-service/src/main/resources/configs/common/application.conf
@@ -87,3 +87,8 @@ scopeFiltersConfig = [
     ]
   }
 ]
+
+log.config = {
+  query.threshold.millis = 1500
+}
+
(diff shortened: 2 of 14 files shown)


gateway-service/src/main/resources/configs/common/application.conf
```diff
@@ -87,3 +87,8 @@ scopeFiltersConfig = [
     ]
   }
 ]
+
+log.config = {
```
surajpuvvada, line 91: do we need to add it to helm templates ?

inespo: Checked helm template, and it has the application.conf as a configmap. So, if it needs to be overriden, we only need to override it in helm-values.yaml.

kotharironak, line 91: Config is specific to EntitySevice, lets move it under the entity service level.
`entity.service.log.config = { query.threshold.millis = 1500}`

And, do we need hierarchy here - query.threashold.millis?
or we can simply have `slowQueryThresholdMs = 1500` ? 
``entity.service.log.config = { slowQueryThresholdMs = 1500}`

inespo: will move to entity.service.log.config. 

The hierarchical naming is more flexible for future expansion, say there's another query logging configuration then, it can follow the same pattern. Also, it's still matches with HOCON convention. Do you see a downside in keeping hierarchical?

inespo: Actually, should it be only for entity.service? Should the other services have access to this configuration as well?

.snyk
```diff
@@ -0,0 +1,9 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JAVA-IOGRPC-571957:
```
kotharironak, line 5: I have already updated this. Can you do resolve it?

PR #8, opened 2020-07-29
fix: aggregations for "The Rest" when multiple groupbys are specified in the request

tim-mwangi:


gateway-service-impl/src/test/resources/expected-test-responses/explore/aggregations-with-groupby-and-the-rest.json
```diff
@@ -0,0 +1,180 @@
+{
```
kotharironak, line 1: @tim-mwangi will this cover the tests for the `rest` filter path?

kotharironak: Ok, I see it will be covered by `*multiple-groupbys*.json`

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -104,32 +103,124 @@ private ExploreRequest createRequest(
...
+    );
+
+    return filterBuilder;
+  }
+
+  private List<String> groupByColumnList(ExploreRequest originalRequest) {
+    return originalRequest
+        .getGroupByList()
+        .stream()
+        .map(groupBy -> groupBy.getColumnIdentifier().getColumnName())
```
kotharironak, line 178: Do we need a filter here `hasColumnIdentifier()`?

tim-mwangi: No need. Only column names are allowed in the group by and the validators check this.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -104,32 +103,124 @@ private ExploreRequest createRequest(
...
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  private Filter.Builder createGroupValuesAndFilter(List<String> groupByColumns, Row.Builder rowBuilder) {
+    Filter.Builder filterBuilder = Filter.newBuilder();
+    filterBuilder.setOperator(Operator.OR);
```
buchi-busireddy, line 184: Shouldn't this be `AND`?

buchi-busireddy: nm. got it.

tim-mwangi: Nope OR. Renamed the method

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -104,32 +103,124 @@ private ExploreRequest createRequest(
...
+  }
+
+  private List<String> getExcludedValues(
       String columnName, ExploreResponse.Builder originalResponse) {
     // GroupBy only supports columns expressions for now.
     return originalResponse.getRowBuilderList().stream()
         .map(rowBuilder -> rowBuilder.getColumnsMap().get(columnName))
         .map(Value::getString)
-        .collect(Collectors.toUnmodifiableSet());
+        .collect(Collectors.toUnmodifiableList());
```
buchi-busireddy, line 220: @tim-mwangi why change to list?

tim-mwangi: It was initially a set because of non-unique group values when multiple group by columns are specified but since have an NOT IN List for each of the groups is a bug - which is what this fix is for - when only one group is requested, we just use a list of the group values since they are unique.

PR #11, opened 2020-07-31
Optimize aggregated metric query if source is the same

inespo:
When the source is the same, optimized the execution tree to
only create 1 SelectionNode that contains both attribute
and aggregated metric data.

As a result, the ExecutionVisitor checks the attributes
and determine to execute it in 1 query or split it to multiple

Full Integration Testing was done to validate the following use cases:
1. Fetch data from same source for attribute and aggregate metrics
2. Fetch data from differetn source for attribute and aggregate metrics
3. Fetch data from same source with filter


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -169,4 +178,10 @@ private QueryNode checkAndAddSortAndPaginationNode(
         executionContext.getEntitiesRequest().getOffset(),
         orderByExpressions);
   }
+
+  private boolean isSingleSourceAndSame(Set<String> firstSource, Set<String> secondSource) {
```
tim-mwangi, line 182: `firstSource.size() == 1 && firstSource.equals(secondSource)`?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -73,48 +74,20 @@ public QueryServiceEntityFetcher(
   @Override
   public EntityFetcherResponse getEntities(
       EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
+    Map<String, AttributeMetadata> attributeMetadataMap =
+        attributeMetadataProvider.getAttributesMetadata(
+            requestContext, AttributeScope.valueOf(entitiesRequest.getEntityType()));
+    // Validate EntitiesRequest
+    entitiesRequestValidator.validate(entitiesRequest, attributeMetadataMap);
```
tim-mwangi, line 81: Can this help you remove some if checks in the code below given that validations will catch the issues?

inespo: Which validation below that you're referring to? I can't find additional validation after this line, mostly extraction.

tim-mwangi: Just wondering if it can simplify some code down below.

inespo: Ah, I see. Mostly is transforming the data. It can be extracted in the next re-factoring.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -98,24 +98,33 @@ QueryNode buildExecutionTree(ExecutionContext executionContext, QueryNode filter
...
+      if (!executionContext.getPendingSelectionSources().isEmpty()) {
+        rootNode =
+            new SelectionNode.Builder(rootNode)
+                .setAttrSelectionSources(executionContext.getPendingSelectionSources())
+                .build();
+        // Handle case where there is no order by but pagination still needs to be done
+        rootNode = checkAndAddSortAndPaginationNode(rootNode, executionContext);
+      }
+
+      if (!executionContext.getPendingMetricAggregationSources().isEmpty()) {
```
tim-mwangi, line 120: These ifs are not exclusive, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
tim-mwangi: This is the same as the one in ExecutionVisitor?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
inespo: I assume you're referring to these 2 ifs: 
```if (!executionContext.getPendingSelectionSources().isEmpty())```
and 
```if (!executionContext.getPendingMetricAggregationSources().isEmpty())``` 
They are not. We need to check for both selection and metric aggregations. Also, nothing changed in the Else section. It's the same as before this refactoring, I just moved it to else condition.

File: gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java (hunk 1 above)
Comment (inespo) [reply gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java:182]: fixed.

File: gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java [hunk 4]
```diff
@@ -141,53 +144,83 @@ public EntityFetcherResponse visit(SelectionNode selectionNode) {
...
-                          request.getEndTimeMillis(),
-                          request.getEntityType(),
-                          executionContext.getRequestHeaders());
-                  return entityFetcher.getAggregatedMetrics(context, request);
-                })
-            .collect(Collectors.toList()));
+    // Select attributes, metric aggregations and time-series data from corresponding sources
+    List<EntityFetcherResponse> resultMapList = new ArrayList<>();
+    // if single-sourced and same source, then combine get Entities and aggregated Metrics
+    if (isSingleSourceAndSame(selectionNode.getAttrSelectionSources(), selectionNode.getAggMetricSelectionSources())) {
```
avinashkolluru, line 150: The ExecutionVisitor should ideally not be aware of the single vs multiple source. All that complex logic should just be in one place and that is in the ExecutionTreeBuilder. 
The decision to use getEntitiesAndAggregatedMetrics  should purely be based on whether the the SelectionNode has both AttrSelectionSources and/or AggMetricSelectionSources. 
That way the code will be very generic.

cc: @tim-mwangi Please see if that is possible and cleanup if you can.

avinashkolluru: Ahhh... sorry... I take that comment back. I misunderstood this a bit.
Please ignore my comment.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -286,51 +237,231 @@ public EntityFetcherResponse getAggregatedMetrics(
...
+              columnValue,
+              attributeMetadataMap);
+        }
+      }
+    }
+    return new EntityFetcherResponse(entityMap);
+  }
+
+  @Override
+  public EntityFetcherResponse getEntitiesAndAggregatedMetrics(
```
avinashkolluru, line 254: @inespo 
I'm trying to understand the need for this method. How is it different from getAggregatedMetrics?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
tim-mwangi: It's possible though if you wanna do it that way, you will have to introduce more QueryNode types. For example EDS does not support aggregations or time aggregations and to support all same source type of queries you will have to either branch in execution visitor or use a new node type if you don't want to branch.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
tim-mwangi: So I think I can echo @inespo thoughts when I say that it may not be that different and could actually be merged into it. But this differentiates the entry points for single source and multiple sources. This whole class needs to be refactored and cleaned up.

PR #16, opened 2020-08-13
fix: (WIP) remove hardcoded limit in explore service requests

tim-mwangi:


tim-mwangi:
Will need to start this from scratch since the code has changed so much. Closing it out.

PR #18, opened 2020-08-17
fix: dont use OR chain when no attribute transformation is needed

tim-mwangi:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -314,13 +315,41 @@ private Filter transformLeafFilter(
       return filter;
     }
 
+    // If the domain object mappings contains one item that equals to the lhs expression column
+    // there's no need to remap. Note equality also means there's no filter in the mapping eg:
+    //    {
+    //      scope = API
```
avinashkolluru, line 321: Just curious. Why would there even be such a mapping?

tim-mwangi: Because we are lazy and co-opted the existing domain configs to also specify well known entity Ids for the entities requests.

PR #20, opened 2020-08-20
Make the request timeout configurable instead of hard-coding them.

buchi-busireddy:


gateway-service-impl/src/test/java/org/hypertrace/gateway/service/GatewayServiceImplTest.java
```diff
@@ -0,0 +1,34 @@
...
+      throw new RuntimeException("Cannot find gateway-service config file in the classpath");
+    }
+
+    Config fileConfig = ConfigFactory.parseFile(new File(configFilePath));
+    appConfig = ConfigFactory.load(fileConfig);
+  }
+
+  @Test
+  public void testServiceImplInitialization() {
+    new GatewayServiceImpl(appConfig);
```
tim-mwangi, line 32: No assertions on this

buchi-busireddy: No failure itself is the pass criteria. There might be a better way to show that though but in this case parsing the config, creating all services, etc will be tested with this.

PR #21, opened 2020-08-21
Reverting the optimization to use DISTINCTCOUNT for entity total query.

buchi-busireddy:
This reverts a minor part of d1659f49898fe3f0c46c24b4d2d359b219ed4ab0. The old behavior is to use groupBy on the entity id and count the records, which is what this commit gets back again.

In the long run, we should get entity total from EDS instead.

### DIFF
--- a/gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
+++ b/gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
@@ -190,25 +190,26 @@ public class EntityServiceTest extends AbstractGatewayServiceTest {
             ).iterator());
 
     // get total request.
-    QueryRequest expectedTotalQueryRequest = QueryRequest.newBuilder()
-        .addAggregation(createQsAggregationExpression("DISTINCTCOUNT", "API.apiId", "DISTINCTCOUNT_entityId_forTotal"))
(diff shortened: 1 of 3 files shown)


gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
```diff
@@ -190,25 +190,26 @@ public void testGetEntitiesOnlySelectFromSingleSource() {
...
 
     // get total request.
-    QueryRequest expectedTotalQueryRequest = QueryRequest.newBuilder()
-        .addAggregation(createQsAggregationExpression("DISTINCTCOUNT", "API.apiId", "DISTINCTCOUNT_entityId_forTotal"))
+    expectedQueryRequest = QueryRequest.newBuilder()
+        .addSelection(createQsColumnExpression("API.apiId")) // Added implicitly in the getEntitiesAndAggregatedMetrics() in order to do GroupBy on the entity id
+        // QueryServiceEntityFetcher adds Count(entityId) to the request for one that does not have an aggregation.
+        // This is because internally a GroupBy request is created out of the entities request and
+        // an aggregation is needed.
+        .addSelection(createQsAggregationExpression("Count", "API.apiId"))
```
tim-mwangi, line 198: What's the timeline to fetch total from EDS?

PR #27, opened 2020-09-04
Add name for the port in service

mohit-a21:
Istio does protocol selection by looking at
the prefix of port name defined in the service.
In case istio is not able to detect the protocol,
it treats the traffic as plain tcp traffic.

### DIFF
--- a/helm/templates/service.yaml
+++ b/helm/templates/service.yaml
@@ -9,6 +9,6 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       targetPort: {{ .Values.containerPortName }}
-      name: grpc-{{ .Chart.Name }}
+      name: {{ .Values.containerPortName }}-{{ .Chart.Name }}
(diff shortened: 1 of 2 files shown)


helm/templates/service.yaml
```diff
@@ -9,5 +9,6 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       targetPort: {{ .Values.containerPortName }}
+      name: grpc-{{ .Chart.Name }}
```
tim-mwangi, line 12: maybe also change this https://github.com/hypertrace/gateway-service/blob/283b78eb77fc73dbbaed1dd6fbaa8df75591059d/helm/values.yaml#L15 to grpc-gateway-service

mohit-a21: Changed to `containerPortName: grpc`.

tim-mwangi (approved):

PR #30, opened 2020-09-17
Fixes default empty string conversion to numerical value

inespo:
Missing attributes will be converted as empty string from EDS Value proto object.
As a result, when a query fetches an optional attribute that could be missing
on some entities, an exception is thrown during conversion.

To avoid throwing an NumerFormatExcpetion, convert empty string to zero
Double / Long data in the Value object, since that would the default value
in proto object as well.


avinashkolluru (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -29,14 +29,30 @@ public Value doConvert(String value, AttributeKind attributeKind, Value.Builder
 
       case TYPE_DOUBLE:
         valueBuilder.setValueType(ValueType.DOUBLE);
-        valueBuilder.setDouble(Double.parseDouble(value));
+        /*
```
avinashkolluru, line 32: Just curious. What happens in the TYPE_BOOL case?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/converter/StringToAttributeKindConverterTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Test
+  public void test_defaultProtoEmptyString_ToNumerical_setToDefault() {
+    StringToAttributeKindConverter converter = StringToAttributeKindConverter.INSTANCE;
+    Value actualValue =
+        converter.doConvert("", AttributeKind.TYPE_DOUBLE, Value.newBuilder());
+    assertEquals(0.0d, actualValue.getDouble());
+
+    actualValue =
+        converter.doConvert("", AttributeKind.TYPE_DOUBLE, Value.newBuilder());
```
avinashkolluru, line 21: Should the type here be TYPE_INT64 instead?

inespo: Yes, fixed this

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
inespo: It would be false by default if the value is NOT "true".

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -29,14 +29,30 @@ public Value doConvert(String value, AttributeKind attributeKind, Value.Builder
...
+        * In proto, by default, field with missing value will be set to empty string. As a result,
+        * optional attributes that could return empty string will cause java.lang.NumberFormatException.
+        * Instead, convert this to 0, which is default value for double and long in proto if
+        * field is missing.
```
jcchavezs, line 36: Thanks for the explanation.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -29,14 +29,30 @@ public Value doConvert(String value, AttributeKind attributeKind, Value.Builder
...
+        */
+        if (value.isEmpty()) {
+          valueBuilder.setDouble(0.0d);
+        } else {
+          valueBuilder.setDouble(Double.parseDouble(value));
```
jcchavezs, line 41: Should we handle the exception on parsing this?

PR #32, opened 2020-10-09
Don't set limit for the entities query when the query to OLAP

buchi-busireddy:
is grouping by multiple columns since that leads to wrong results.
For now we set limit to a higher value in those cases.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -273,9 +273,21 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
     QueryRequest.Builder builder =
         constructSelectionQuery(requestContext, entitiesRequest, entityIdAttributes, aggregates);
 
-    // Order by, limit and offset from the request.
-    builder.setLimit(entitiesRequest.getLimit());
-    builder.setOffset(entitiesRequest.getOffset());
+    // If there is more than one groupBy column, we cannot set the same limit that came
+    // in the request since that might return less entities than needed when the same
+    // entity has different values for the other group by columns. Example: A service entity's
+    // name changes and that will now have two different names.
+    // For now, we pass a high value of limit in this case so that we get all the entities.
+    // Limit has to be applied post the query in this case. Setting offset also might be wrong
+    // here, hence not setting it.
+    if (builder.getGroupByCount() > 1) {
+      builder.setLimit(QueryServiceClient.DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+    } else {
+      builder.setLimit(entitiesRequest.getLimit());
+      builder.setOffset(entitiesRequest.getOffset());
+    }
+
+    // Order by from the request.
     builder.addAllOrderBy(
             QueryAndGatewayDtoConverter.convertToQueryOrderByExpressions(entitiesRequest.getOrderByList()));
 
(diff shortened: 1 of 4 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -273,9 +273,21 @@ public EntityFetcherResponse getEntitiesAndAggregatedMetrics(
...
-    builder.setOffset(entitiesRequest.getOffset());
+    // If there is more than one groupBy column, we cannot set the same limit that came
+    // in the request since that might return less entities than needed when the same
+    // entity has different values for the other group by columns. Example: A service entity's
+    // name changes and that will now have two different names.
+    // For now, we pass a high value of limit in this case so that we get all the entities.
+    // Limit has to be applied post the query in this case. Setting offset also might be wrong
+    // here, hence not setting it.
+    if (builder.getGroupByCount() > 1) {
+      builder.setLimit(QueryServiceClient.DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
```
tim-mwangi, line 284: Can you check the ExecutionVisitor and see if you need to do the limit and offset yourself?

buchi-busireddy: Yeah I've checked that already. There, we just copy the limit that comes in the request. This is more of OLAP specific implementation detail, needn't spread outside QSEFetcher class.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
```diff
@@ -191,7 +193,6 @@ public void test_getTotalEntitiesSingleEntityIdAttribute() {
             .setEntityType(entityType.name())
             .setStartTimeMillis(startTime)
             .setEndTimeMillis(endTime)
-            .addSelection(buildExpression(API_NAME_ATTR))
```
tim-mwangi, line 194: Why was this removed?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -354,6 +346,22 @@ public EntityFetcherResponse getEntitiesAndAggregatedMetrics(
     return new EntityFetcherResponse(entityBuilders);
   }
 
+  private void adjustLimitAndOffset(QueryRequest.Builder builder, int limit, int offset) {
+    // If there is more than one groupBy column, we cannot set the same limit that came
+    // in the request since that might return less entities than needed when the same
+    // entity has different values for the other group by columns. Example: A service entity's
+    // name changes and that will now have two different names.
```
avinashkolluru, line 353: Based on this example, I think its more a problem in having mutable fields(like api_name, service_name) in pinot
If that is the only case we are trying to solve, may be a better(but requiring effort) approach is to not add the api_name in the group by clause and implement a custom aggregation function like a LATEST(not entirely sure if this is possible too)

Based on https://docs.pinot.apache.org/developers/developers-and-contributors/extending-pinot/custom-aggregation-function, it looks possible, but needs more research

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
buchi-busireddy: Hmm.. was trying to modify this test case to cover both use cases but not needed so reverted it.

tim-mwangi (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
buchi-busireddy: We discussed about this with @aaron-steinfeld and @tim-mwangi This was one of the ideas I had too and I've even asked in Pinot community about the feasibility of `LATEST()`. See https://apache-pinot.slack.com/archives/CDRCA57FC/p1602265559126800

I'll log a story to continue the research.

PR #33, opened 2020-10-09
fix: converting string to string list for TYPE_STRING_ARRAY attribute kind

tim-mwangi:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -14,7 +15,8 @@
 
 public class StringToAttributeKindConverter extends ToAttributeKindConverter<String> {
   private static final Logger log = LoggerFactory.getLogger(StringToAttributeKindConverter.class);
```
buchi-busireddy, line 17: Nit. `s/log/LOGGER/`

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -69,20 +71,38 @@ public Value doConvert(String value, AttributeKind attributeKind, Value.Builder
...
+  private List<String> convertToArray(String jsonString) {
+    if (StringUtils.isEmpty(jsonString)) {
+      return List.of();
+    }
+
+    try {
+      return objectMapper.readValue(jsonString, LIST_TYPE_REFERENCE);
+    } catch (IOException e) {
+      log.warn("Unable to read List JSON String data from: {}. With error:", jsonString, e);
+      log.warn("Setting data as empty list instead");
```
buchi-busireddy, line 93: Do we need the second warning? it should be implicit or merge it with the above line itself.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
tim-mwangi: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
tim-mwangi: Will merge into the other log message.

buchi-busireddy (approved):

PR #35, opened 2020-10-19
fix: allow for mutations across domains

tim-mwangi:
(wip)
This is dropped now since we have decided not to have labels for domains.


tim-mwangi:
Closing this one out. Also it's not possible to do since domains are a query time concept and span across API and Service scopes. So which entity labels would be modified?

PR #37, opened 2020-10-26
Removing AttributeScope enum.

buchi-busireddy:
This is needed for us to be able to add entities dynamically without code changes.


gateway-service-impl/build.gradle.kts
```diff
@@ -28,12 +34,19 @@ dependencies {
   // Metrics
   implementation("io.dropwizard.metrics:metrics-core:4.1.9")
 
+  runtimeOnly("io.netty:netty-codec-http2:4.1.53.Final") {
```
aaron-steinfeld, line 37: Does the impl use netty or is it only needed by the service (which also declares this)?

aaron-steinfeld (dismissed):

gateway-service-impl/build.gradle.kts
buchi-busireddy: I don't think impl should use netty but I was surprised to see this failure where impl module was reporting Snyk failure in these libraries. https://app.circleci.com/pipelines/github/hypertrace/gateway-service/195/workflows/528c60c9-0589-49e5-848d-9f486838ed25/jobs/564/parallel-runs/0/steps/0-106

I don't see that dependency coming from anywhere else too. Did you see such behavior?

aaron-steinfeld: I haven't run into that in an impl package. I checked, netty is not part of the impl's runtime classpath. I'd guess line 51 triggered it since I believe snyk scans all configurations by default. So I'd suggest two things: change 51 to `testRuntimeOnly` (if it's actually needed, most unit tests don't unless you're actually spinning up a local channel which is usually not a great idea) , and move the constraints to testRuntimeOnly as well.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -358,7 +358,7 @@ private void addInteractions(
                 QueryRequestUtil.createBetweenTimesFilter(
                     metadataProvider
                         .getAttributeMetadata(
-                            requestContext, AttributeScope.INTERACTION, TIMESTAMP_COLUMN_KEY)
+                            requestContext, AttributeScope.INTERACTION.name(), TIMESTAMP_COLUMN_KEY)
```
tim-mwangi, line 361: Why are we still using INTERACTION enum?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
```diff
@@ -153,9 +151,7 @@ public static AttributeMetadata getAttributeMetadata(
 
   private static AttributeMetadata getAttributeMetadata(
       AttributeMetadataProvider attributeMetadataProvider,
-      RequestContext requestContext,
-      AttributeScope scope,
-      String key) {
+      RequestContext requestContext, String scope, String key) {
```
tim-mwangi, line 154: mmh I prefer line by line continuation

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
```diff
@@ -60,7 +60,7 @@ public SpansResponse getSpansByFilter(RequestContext context, SpansRequest reque
     final Context timerContext = queryExecutionTimer.time();
     try {
       Map<String, AttributeMetadata> attributeMap =
-          attributeMetadataProvider.getAttributesMetadata(context, AttributeScope.EVENT);
+          attributeMetadataProvider.getAttributesMetadata(context, AttributeScope.EVENT.name());
```
tim-mwangi, line 63: ditto about the enum

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
buchi-busireddy: Yes, because this doesn't come from the user request. This is implicit in the entities request when user asks for incoming and outgoing edges right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
buchi-busireddy: Hmm.. interestingly, java code conventions plugin doesn't seem to lean towards one approach. Do you want me to fix this?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
buchi-busireddy: Yeah this service itself is specific to spans, hence this enum is hard-coded. This doesn't come from user request directly.

gateway-service-impl/build.gradle.kts
buchi-busireddy: Thanks. That seems to be the case. i've fixed this now.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
tim-mwangi: So it seems to me that we are moving away completely from AttributeScope enum and this is hardcoded over here. Why can't we use a string const?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
tim-mwangi: sure.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
tim-mwangi: See the comment above about moving away completely from AttributeScope enum.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
buchi-busireddy: Done.

tim-mwangi (approved):

aaron-steinfeld (approved):

PR #39, opened 2020-11-02
Fixing a bug in merging the results from different data sources in ExecutionVisitor.

buchi-busireddy:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -226,7 +226,12 @@ public EntityFetcherResponse visit(SelectionNode selectionNode) {
                   return entityFetcher.getTimeAggregatedMetrics(requestContext, request);
                 })
             .collect(Collectors.toList()));
-    return resultMapList.stream().reduce(result, (r1, r2) -> union(Arrays.asList(r1, r2)));
+    // Union all the responses from this single source first.
+    EntityFetcherResponse currentResponse =
```
avinashkolluru, line 230: Nice!

avinashkolluru: How did we find this?

avinashkolluru, line 230: Little confused. The expectation is that the `filter` in line 156 is constructed from the childNodeResponse.
So why do we need to intersect again?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -142,18 +142,18 @@ public EntityFetcherResponse visit(OrNode orNode) {
...
+    EntityFetcherResponse childNodeResponse = selectionNode.getChildNode().acceptVisitor(this);
 
     // If the result was empty when the filter is non-empty, it means no entities matched the filter
     // and hence no need to do any more follow up calls.
-    if (result.isEmpty()
+    if (childNodeResponse.isEmpty()
         && !Filter.getDefaultInstance().equals(executionContext.getEntitiesRequest().getFilter())) {
       LOG.debug("No results matched the filter so not fetching aggregate/timeseries metrics.");
-      return result;
+      return childNodeResponse;
```
tim-mwangi, line 152: no coverage for this line.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
buchi-busireddy: Because the first source might have more entities that what would be matched in the second source. For example, if you query is something like "Get APIs whose apiDiscoveryState is DISCOVERED and latency > 100ms", you might have 10 APIs matching first clause in EDS but only have two matching the second clause in QS. The final result should only have two APIs but the current code returns all 10 APIs, because we're doing a union.

Having said, this code needs to support `OR`'ing the query b/w multiple data sources in future as well but that's next enhancement.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
buchi-busireddy: Added coverage for this case as well now.

tim-mwangi (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
avinashkolluru: @buchi-busireddy The example you gave is the functionality of the AndNode. SelectionNode is where you've already filtered the entity ids based on the filter conditions and you are only selecting additional attributes.

avinashkolluru: I'm inclined to think we've fundamentally broken then execution tree that was built and thats why we ended up in this scenario. 
Any specific issue/query that resulted in this change?
cc: @tim-mwangi

tim-mwangi: I don't think so. I mean the only change that's major is making another route for single source requests which I kept totally separate from the multiple source requests because I did want to complicate the complexity further. I think this has always been an issue. The only thing that I can think of is graphql refactor which exposed this. Graphql now craps out when it gets back results missing some of the column values requested for.

buchi-busireddy: Agree with @tim-mwangi I think this has always been there.
@avinashkolluru The bug is more around response merging. AndNode is just delegating to the visitor. ExecutionVisitor is the main one which has key logic around how execution is done and responses are merged, which is why the fix is in that class.

avinashkolluru: The ExecutionVisitor is the one which visits each node. I meant that in the visit(AndNode) is where we need to do the intersect based on the example you gave.

The DataFetcherNodes are leaf nodes which apply the filters and fetch the results. If there query filter spans 2 different source, then there should be 2 of these DataFetcherNodes and their parent will be a AndNode. The visit(AndNode) takes care of inserting the results because its an AND condition(If it was a OR, we do a union)

SelectionNode comes later, when we done all the filtering and got the entity ids, but now we need to select additional fields from the different sources. So ideally, the visit(SelectNode) should never filter out and entires returned by its children.

avinashkolluru: If GraphQL craps out when it is missing column values(probably because EDS doesn't have the value, but QS has it) then that doesn't mean we should just filter the results and send the entities that have all the values. There has to be a different way to handle that.
By doing an intersect across sources, we are just getting past that problem(that graphql can't handle), but erroneously.

buchi-busireddy: @avinashkolluru See the below query. `DataFetcherNode` needn't be the leaf always and in some cases the `SelectNode` itself is a leaf which is fetching the results. `visit(AndNode)` is already intersecting so no issue there.

```
SelectionNode{attrSelectionSources=[QS], aggMetricSelectionSources=[], timeSeriesSelectionSources=[], 
childNode=SortAndPaginateNode{limit=500, offset=0, orderByExpressionList=[expression {
  function {
    function: PERCENTILE
    arguments {
      columnIdentifier {
        columnName: "SERVICE.duration"
      }
    }
    arguments {
      literal {
        value {
          valueType: LONG
          long: 99
        }
      }
    }
    alias: "PERCENTILE_SERVICE.duration_[99]"
  }
}
order: DESC], 
childNode=SelectionNode{attrSelectionSources=[], aggMetricSelectionSources=[QS], timeSeriesSelectionSources=[], 
childNode=DataFetcherNode{source='EDS', filter=lhs {
  columnIdentifier {
    columnName: "SERVICE.labels"
  }
}
operator: EQ
rhs {
  literal {
    value {
      valueType: STRING
      string: "Critical"
    }
  }
}
}}}}
```

Taking the above query as an example, while merging results from different parts of the tree, we must intersect because we're passing the ids from one side as filter to the other side. This is functional issue, we aren't fixing this to satisfy GraphQL or other layers.

PR #41, opened 2020-11-04
Fixing a bug in merging data from multiple sources in query execution visitor.

buchi-busireddy:
Currently, though query filter between two different data sources is `AND`ed, the results are unioned. This PR fixes that by converting them into intersection.


buchi-busireddy:
FTR, we have discussed this and, though this approach fixes the issue, the right approach is to treat timestamp also like any other attribute and bring that into source based filters. I'll try to fix in that approach, hence closing this.

PR #44, opened 2020-11-11
Timestamp filter refactoring to fix time range filtering bugs in entity queries.

buchi-busireddy:
**Why this PR?**
This PR is fixing a bug in entity queries, which causes the query layer to error out when a query is asking for entity attributes that span across the EDS and QueryService data sources, in the case there is no data for that entity in QueryService.

**What's the fix?**
Currently, the query federation code that we have (see https://github.com/hypertrace/gateway-service/tree/main/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query) doesn't treat the timestamp predicates in the query like other regular predicates. The timestamps are special cased and they're added to the upstream requests only when the query is being sent to a remote service. That's what caused the above mentioned bug.

This PR treats the timestamp predicates, just like any other predicates in the filter. However, the timestamps are first-class fields in the API and we don't want to remove them because they're mandatory in each entity query (at least in the current design). Hence, the timestamp filter is added to the query request in the request pre-processor so that all the query execution tree shouldn't special case timestamp. Code simplified in those layers and even the implementations of `IEntityFetcher` interface.

There is another existing tech debt (https://github.com/hypertrace/hypertrace/issues/119), which this PR isn't addressing yet. We current use the object received on the API (the proto class `EntitiesRequest`) throughout in the implementation code but we shouldn't. That will be a nice tech debt to cleanup when we get time.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -285,6 +285,11 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
       LOG.debug("Sending Query to Query Service ======== \n {}", queryRequest);
     }
 
+    try {
+      System.out.println("Actual: " + JsonFormat.printer().omittingInsignificantWhitespace().print(queryRequest));
+    } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
+    }
     Iterator<ResultSetChunk> resultSetChunkIterator =
         queryServiceClient.executeQuery(queryRequest, requestContext.getHeaders(), requestTimeout);
 
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -724,8 +724,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
   }
 
   /**
-   * - Adds the time range to the filter - Adds a non null filter on entity id - Converts it to the
-   * query service filter
+   * Converts the filter in the given request to the query service filter and adds a non null filter on entity id.
    */
   private Filter.Builder constructQueryServiceFilter(EntitiesRequest entitiesRequest, List<String> entityIdAttributes) {
     Filter.Builder filterBuilder = convertToQueryFilter(entitiesRequest.getFilter());
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -1,6 +1,7 @@
 package org.hypertrace.gateway.service.common.datafetcher;
 
 import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDtoConverter.convertToQueryExpression;
+import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDtoConverter.convertToQueryFilter;
 
 import com.google.common.base.Preconditions;
 import java.time.Duration;
@@ -13,6 +14,9 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
 import org.apache.commons.lang3.StringUtils;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.query.service.api.ColumnMetadata;
@@ -99,6 +103,11 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
       LOG.debug("Sending Query to Query Service ======== \n {}", queryRequest);
     }
 
+    try {
+      System.out.println("actual: " + JsonFormat.printer().omittingInsignificantWhitespace().print(queryRequest));
+    } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
+    }
     Iterator<ResultSetChunk> resultSetChunkIterator =
         queryServiceClient.executeQuery(queryRequest, requestContext.getHeaders(),
             requestTimeout);
@@ -371,7 +380,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
             .map(Expression.Builder::build)
             .collect(Collectors.toList());
     Filter.Builder filterBuilder =
-        constructQueryServiceFilter(entitiesRequest, requestContext, entityIdAttributes);
+        constructQueryServiceFilter(entitiesRequest, entityIdAttributes);
 
     QueryRequest.Builder builder =
         QueryRequest.newBuilder()
@@ -704,7 +713,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
 
     Filter.Builder queryFilter =
         constructQueryServiceFilter(
-            timeAlignedEntitiesRequest, timeAlignedEntitiesRequestContext, idColumns);
+            timeAlignedEntitiesRequest, idColumns);
     builder.setFilter(queryFilter);
 
     // First group by the id columns.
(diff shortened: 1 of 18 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -109,10 +110,23 @@ public EntitiesRequest transform(
         transformFilter(attributeIdMappings, originalRequest.getFilter(), context);
     Filter filter = mergeFilters(filterToInject, transformedFilter);
 
+    // Convert the time range into a filter and set it on the request so that all downstream
```
buchi-busireddy, line 113: @tim-mwangi @aaron-steinfeld @avinashkolluru This is the change I think fixes the bug around entity filtering when time range is involved and also keeps the query execution tree code simpler. 
Wanted to get your inputs on this before I go deep down this approach.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -99,6 +103,11 @@ public EntityFetcherResponse getEntities(
       LOG.debug("Sending Query to Query Service ======== \n {}", queryRequest);
     }
 
+    try {
```
buchi-busireddy, line 106: was debugging with this. will remove later.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
aaron-steinfeld: I don't know what all goes on in the query execution tree as to why this is actually needed, but my concern would be that there are now two sources of truth WRT the time range. Code that needs to directly work with the time range (for example, the bucket alignment code, if that's still needed), would need to either read from the first class field, or add significant complexity to read from the filter tree.

Maybe it would help if you could show where the code gets simpler as a result of this?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
```diff
@@ -162,29 +175,31 @@ public static Filter createQsDefaultRequestFilter(String timestampColumnName,
...
             Filter.newBuilder()
+                .setOperator(Operator.AND)
                 .addChildFilter(
                     createQsFilter(
-                        createQsColumnExpression(entityIdColumnName),
-                        Operator.NEQ,
-                        createQsStringLiteralExpression("null")
+                        createQsColumnExpression(timestampColumnName),
+                        Operator.GE,
+                        createQsLongLiteralExpression(startTime)
```
tim-mwangi, line 186: question: QS now works nicely with types?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
```diff
@@ -162,29 +175,31 @@ public static Filter createQsDefaultRequestFilter(String timestampColumnName,
...
                 )
-        ).build();
+                .addChildFilter(
+                    createQsFilter(
+                        createQsColumnExpression(timestampColumnName),
+                        Operator.LT,
+                        createQsLongLiteralExpression(endTime)
+                    )
+                ))
+        .addChildFilter(
```
tim-mwangi, line 196: Is this added twice?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -109,10 +110,23 @@ public EntitiesRequest transform(
...
+    // components needn't treat it specially.
+    String timestampAttributeId = AttributeMetadataUtil.getTimestampAttributeId(
+        attributeMetadataProvider, context, originalRequest.getEntityType());
+
+    Filter.Builder filterBuilder = Filter.newBuilder()
```
tim-mwangi, line 118: If we are doing this here, then have you removed the code to convert the default filter in QueryServiceRequestAndResponseUtils.createQsDefaultRequestFilter()?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -109,10 +110,23 @@ public EntitiesRequest transform(
...
+        .setOperator(AND)
+        .addChildFilter(getTimestampFilter(timestampAttributeId, Operator.GE, originalRequest.getStartTimeMillis()))
```
tim-mwangi, line 120: Also if you do this, what about requests going to EDS? They will have time range in the filter as well and that might cause issues since those columns dont map to anything in entity attributes.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
buchi-busireddy: Oh gosh.. I didn't know that but would verify that it does otherwise i'll have to revert this change :)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: Timestamp columns will only have QS as the source so they'd never go to EDS.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
buchi-busireddy: No, i see it only once.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
aaron-steinfeld: I can't remember either, but looking at the old filter building code, it was working in strings so I'd suspect we'd need to do the same.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
tim-mwangi: But at this point you do not know where the query will go to. And if this filter hits the ExecutionTree then it ends up being split into two even if all the columns can be sourced from EDS. Because the time range is the filter now another query is made to QS just to get ids. And then the join happens and I'm not sure about this but you might return empty results since there's nothing from QS when there's no data?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
tim-mwangi: Might be unrelated to this change but this exists in QueryServiceEntityFetcher.constructQueryServiceFilter() as well. Which is also coincidentally the method that adds time range to the query service request.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
aaron-steinfeld: I think this is a separate but important question around liveness - even if all selections are time agnostic, shouldn't the returned entities be the same whether or not a metric is requested? In other words, I'd say this behavior was a bug, and we should never serve a response solely from ES.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
buchi-busireddy: Correct. That will now be removed since time range will not be special case anymore with this change.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: That change is on the way. I wanted to raise the draft with key change to get consensus first.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilderTest.java
```diff
@@ -411,31 +419,40 @@ public void testExecutionTreeBuilderWithSelectFilterOrderPagination() {
...
-      assertTrue(executionTree instanceof SelectionNode);
-      assertTrue(
-          ((SelectionNode) executionTree)
-              .getAttrSelectionSources()
-              .contains(AttributeSource.QS.name()));
-      QueryNode firstChild = ((SelectionNode) executionTree).getChildNode();
-      assertTrue(firstChild instanceof SortAndPaginateNode);
-      QueryNode grandchild = ((SortAndPaginateNode) firstChild).getChildNode();
-      assertEquals(AttributeSource.EDS.name(), ((DataFetcherNode) grandchild).getSource());
+      System.out.println(executionTree);
```
tim-mwangi, line 442: remove println

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
tim-mwangi: In another PR?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: Updated the PR description with details and also logged https://github.com/hypertrace/hypertrace/issues/119 to fix the tech-debt to clear confusion.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
buchi-busireddy: I've tested this and it works fine.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/QueryServiceRequestAndResponseUtils.java
buchi-busireddy: This is already fixed in this PR.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: Fixed this. Please check now.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: Yes, liveness needs to be addressed separately but @tim-mwangi what you're describing should be the current behavior. As long as there is time range in the query, we should return entities only if they had data in that time range. And, that will be maintained with this change.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilderTest.java
buchi-busireddy: Done.

tim-mwangi (approved):

PR #45, opened 2020-11-18
fix: Pushing down the mandatory time range check from entities request to QueryServiceEntityFetcher.

buchi-busireddy:
## Description
Pushing down the mandatory time range check from entities request to QueryServiceEntityFetcher, because time range is mandatory only for query service.

This allows the entity queries which only require data from EDS to work without any
time range.

### Testing
Apart from the unit tests, tested with locally built image in real setup.

### Checklist:

### Documentation
The start_time_millis and end_time_millis in the `EntitiesRequest` aren't mandatory anymore with this change,
unless you are either requesting attributes that are changing with time or asking for metrics. As an example, all you
need is api name, id and http methods then the query can hit EDS and no need of time range.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
@@ -169,12 +169,6 @@ public class GatewayServiceImpl extends GatewayServiceGrpc.GatewayServiceImplBas
       Preconditions.checkArgument(
           request.getSelectionCount() > 0, "Selection list can't be empty in the request.");
 
-      Preconditions.checkArgument(
-          request.getStartTimeMillis() > 0
-              && request.getEndTimeMillis() > 0
-              && request.getStartTimeMillis() < request.getEndTimeMillis(),
-          "Invalid time range. Both start and end times have to be valid timestamps.");
-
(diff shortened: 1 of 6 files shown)


aaron-steinfeld:
So I started to say this offline, but I pretty strongly feel that the time range should remain part of the request validation. In our existing time-ranged queries, the returned entities are the ones with data in the time range. So IMO a query must always go to QS to ensure it has data and the existing behavior was a bug. We should never have query behavior where selecting different sets of attributes returns different sets of entities depending on the time-awareness of the data sources. If we want to support queries without time ranges, which I'm in support of for a number of reasons (for example, joining entities to other queries), I think that's fine but it should be a new API that doesn't support time ranges, metrics or time series.

buchi-busireddy:
@tim-mwangi what are your thoughts on separate dedicated API without time range? For the use cases that you've seen, do you think that would work or would the API be very close to the existing API?

tim-mwangi:
> @tim-mwangi what are your thoughts on separate dedicated API without time range? For the use cases that you've seen, do you think that would work or would the API be very close to the existing API?

Sure it should work and should also always be wired to EDS and validate that the selections, filtering and ordering are only on attributes and never metrics. My concern is time to put it in and resources to get there.

buchi-busireddy:
Closing this PR. Will add this functionality in a new API as we discussed here.

PR #46, opened 2020-11-21
fix: Fixing a bug in query execution, which wasn't passing the time range in filter to query service.

buchi-busireddy:
## Description
This is fixing a bug where the time range wasn't being passed to query service in the query service requests. You can say this was a bug introduced as part of https://github.com/hypertrace/gateway-service/pull/44 bug fix.

### Testing
Unit testing, tested in real setup and verified that all the query service queries have time range now.

### Checklist:

### Documentation
No functional changes and bug was introduced in this release itself.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -4,8 +4,6 @@ import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDt
 import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDtoConverter.convertToQueryFilter;
 
 import com.google.common.base.Preconditions;
-import com.google.protobuf.InvalidProtocolBufferException;
-import com.google.protobuf.util.JsonFormat;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.Collection;
@@ -279,11 +277,6 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
       LOG.debug("Sending Query to Query Service ======== \n {}", queryRequest);
     }
 
-    try {
-      System.out.println("Actual: " + JsonFormat.printer().omittingInsignificantWhitespace().print(queryRequest));
-    } catch (InvalidProtocolBufferException e) {
-      e.printStackTrace();
-    }
     Iterator<ResultSetChunk> resultSetChunkIterator =
         queryServiceClient.executeQuery(queryRequest, requestContext.getHeaders(), requestTimeout);
 
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -4,6 +4,8 @@ import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDt
 import static org.hypertrace.gateway.service.common.converters.QueryAndGatewayDtoConverter.convertToQueryFilter;
 
 import com.google.common.base.Preconditions;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.Collection;
@@ -14,7 +16,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
-
 import org.apache.commons.lang3.StringUtils;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.query.service.api.ColumnMetadata;
@@ -278,6 +279,11 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
       LOG.debug("Sending Query to Query Service ======== \n {}", queryRequest);
     }
 
+    try {
+      System.out.println("Actual: " + JsonFormat.printer().omittingInsignificantWhitespace().print(queryRequest));
+    } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
+    }
     Iterator<ResultSetChunk> resultSetChunkIterator =
         queryServiceClient.executeQuery(queryRequest, requestContext.getHeaders(), requestTimeout);
 
@@ -373,7 +379,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
             .map(Expression.Builder::build)
             .collect(Collectors.toList());
     Filter.Builder filterBuilder =
-        constructQueryServiceFilter(entitiesRequest, entityIdAttributes);
+        constructQueryServiceFilter(entitiesRequest, requestContext, entityIdAttributes);
 
     QueryRequest.Builder builder =
         QueryRequest.newBuilder()
@@ -531,7 +537,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
       long periodSecs = Duration.of(period.getValue(), unit).getSeconds();
       QueryRequest request =
           buildTimeSeriesQueryRequest(
-              entitiesRequest, periodSecs, batch, idColumns, timeColumn);
+              entitiesRequest, requestContext, periodSecs, batch, idColumns, timeColumn);
 
       if (LOG.isDebugEnabled()) {
         LOG.debug(
@@ -674,6 +680,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
 
   private QueryRequest buildTimeSeriesQueryRequest(
       EntitiesRequest entitiesRequest,
+      EntitiesRequestContext context,
       long periodSecs,
       List<TimeAggregation> timeAggregationBatch,
       List<String> idColumns,
@@ -696,9 +703,7 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
             builder.addSelection(
                 QueryAndGatewayDtoConverter.convertToQueryExpression(e.getAggregation())));
 
-    Filter.Builder queryFilter =
-        constructQueryServiceFilter(
-            timeAlignedEntitiesRequest, idColumns);
+    Filter.Builder queryFilter = constructQueryServiceFilter(timeAlignedEntitiesRequest, context, idColumns);
     builder.setFilter(queryFilter);
 
     // First group by the id columns.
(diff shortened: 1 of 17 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -740,6 +744,30 @@ private QueryRequest buildTimeSeriesQueryRequest(
...
                 .collect(Collectors.toList()));
+
+    // Convert the existing filter into query service filter.
+    Filter queryFilter = convertToQueryFilter(entitiesRequest.getFilter()).build();
+    if (!Filter.getDefaultInstance().equals(queryFilter)) {
+      filterBuilder.addChildFilter(queryFilter);
+    }
+
+    // Time range is a mandatory filter for query service, hence add it if it's not already present.
+    if (!hasTimeRangeFilter(queryFilter, context.getTimestampAttributeId())) {
```
buchi-busireddy, line 755: This is the actual bug fix, where if the timestamp filter isn't already present, we'll explicitly add it here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -76,12 +76,20 @@ public EntitiesRequest transform(
         AttributeMetadataUtil.getAttributeIdMappings(
             attributeMetadataProvider, context, originalRequest.getEntityType());
     EntitiesRequest.Builder entitiesRequestBuilder = EntitiesRequest.newBuilder(originalRequest);
+
+    // Convert the time range into a filter and set it on the request so that all downstream
```
buchi-busireddy, line 80: Just code cleanup, no functional change.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntitiesRequestContext.java
```diff
@@ -6,32 +6,40 @@
 
 public class EntitiesRequestContext extends QueryRequestContext {
   private final String entityType;
+  private final String timestampAttributeId;
```
buchi-busireddy, line 9: Instead of recomputing the timestamp column for a given entity over and over, we compute once and carry in the context.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
jcchavezs: Thanks for pointing out. I wonder if there should be a case showing this bug in QueryServiceEntityFetcherTests.java?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
```diff
@@ -275,6 +266,27 @@ public void testGetEntitiesOnlySelectFromMultipleSources() {
             .addSelection(getExpressionFor("API.apiName", "API Name"))
             .addSelection(getExpressionFor("API.httpMethod", "API Http method"))
             .build();
+    QueryRequest expectedQueryRequest = QueryRequest.newBuilder()
+        .setFilter(
+            QueryServiceRequestAndResponseUtils.createQsDefaultRequestFilter("API.startTime",
```
buchi-busireddy, line 271: This filter always has a time range filter in it, there by verified that the query request has time range filter.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
buchi-busireddy: Good ask. There was a test but the assertion wasn't verifying the complete query service request. Added that now at https://github.com/hypertrace/gateway-service/pull/46/files#diff-8fa94e86f08f72f2fe9bd4fbe8a691758efd65b10513050e7ca568ffbb8f7098R269

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -222,6 +226,7 @@ public EntityFetcherResponse visit(SelectionNode selectionNode) {
                           request.getStartTimeMillis(),
                           request.getEndTimeMillis(),
                           request.getEntityType(),
+                          executionContext.getTimestampAttributeId(),
```
tim-mwangi, line 229: Nit: Not covered in tests

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
tim-mwangi: codecov diff still shows this if branch is not covered.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessorTest.java
```diff
@@ -56,16 +56,18 @@ public void teardown() {
...
+        new EntitiesRequestContext(TEST_TENANT_ID, startTime, endTime, "SERVICE", "SERVICE.startTime", Map.of());
     mockAttributeMetadata(entitiesRequestContext, AttributeScope.SERVICE.name(), "id");
     mockAttributeMetadata(entitiesRequestContext, AttributeScope.SERVICE.name(), "startTime");
 
     EntitiesRequest entitiesRequest =
         EntitiesRequest.newBuilder()
             .setEntityType("SERVICE")
-            .setStartTimeMillis(0L)
-            .setEndTimeMillis(1L)
+            .setStartTimeMillis(startTime)
```
tim-mwangi, line 69: What's wrong with 0 and 1

tim-mwangi:
@buchi-busireddy if you are confident with this fix, we can merge it and fast track its promotion to prod. We can put in the unit tests later

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
buchi-busireddy: Please check now @tim-mwangi

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessorTest.java
buchi-busireddy: Nothing wrong. I mainly wanted to pull out the start and end times into variables so that I can reuse and I just assigned real timestamp values to them while I was there.

buchi-busireddy:
> @buchi-busireddy if you are confident with this fix, we can merge it and fast track its promotion to prod. We can put in the unit tests later

Done. More unit tests added now. @tim-mwangi please check.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
buchi-busireddy: Check now.. it's all green :)

tim-mwangi: ok.

tim-mwangi (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -740,6 +737,41 @@ private QueryRequest buildTimeSeriesQueryRequest(
...
+      filterBuilder.addChildFilter(Filter.newBuilder().setOperator(Operator.AND)
+          .addChildFilter(createTimeFilter(context.getTimestampAttributeId(), Operator.GE, entitiesRequest.getStartTimeMillis()))
+          .addChildFilter(createTimeFilter(context.getTimestampAttributeId(), Operator.LT, entitiesRequest.getEndTimeMillis())));
+    }
+
+    return filterBuilder;
+  }
+
+  private static Filter createTimeFilter(String columnName, Operator op, long value) {
```
avinashkolluru, line 757: nit: Why not use TimeRangeFilterUtil here?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -76,12 +76,20 @@ public EntitiesRequest transform(
...
+    // components needn't treat it specially.
+    Filter filter = TimeRangeFilterUtil.addTimeRangeFilter(
```
avinashkolluru, line 82: Just thinking out loud:
An alternative approach to this could be to only to modify the ExecutionTreeBuilder.buildFilterTree to take in the EntitiesRequest and construct the DataFetcherNode for the timestamp condition too.
That way all of this preprocessing and and special casing in the QueryServiceEntityFetcher can probably be avoided. i.e. they continue to work with the first class time fields in the EntitiesRequest.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
buchi-busireddy: That's gateway-service filter vs query service filter. Thought about adding another one but naming is hard :)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
buchi-busireddy: @avinashkolluru That's already being done. That was my previous PR :) However, the non-leaf nodes don't take the time range filter because the only filter they apply is based on child node's results. That ended up removing time range filter for metric queries and time series queries, which is what caused another bug, to which this PR is a fix.

Here is the latest status:
* Time range is added to the request filters in the preprocessor
* Execution tree creates a separate DataFetcherNode for this
* QueryServiceEntityFetcher checks if there is time range filter, adds it if not. I'm not really sure this logic is clean so, if possible, I'd like to fix this cleanly in the next iteration.

avinashkolluru: I think i wasn't clear or may there is some disconnect. Here is what i was thinking:
1) Only modify the buildFilterTree to construct and extra DataFetcherNode based on the start/end time in the EntitiesRequest
2) The non leaf nodes i.e. SelectionNode, continues to work how it was earlier, where the EntitiesRequest filter(in ExecutionVisitor.accept()) is only based on its child nodes results. But, internally in QueryServiceDataFetcher, we anyways use the start/end time from the EntitiesRequest and construct the appropriate query-service request. Basically, all of this doesn't change at all.

So, just by doing 1), I think we would have solved the original problem where the initial request to query-service was going without the time range(because the new DataFetcherNode based on the time range will take care of it)

buchi-busireddy: Thanks for clarifying and I see the approach now. IMO, hard-coding time range filter in QueryServiceEntityFetcher isn't clean and hence I was thinking of approaches to get rid of that instead. Let me think a bit more on this.

PR #51, opened 2020-11-30
feat: delete domain configs and refactor code to accomodate entity id columns config

tim-mwangi:
## Description
Remove obsolete domain configs and switch to using entity id columns config to inform the service the primary key id column names for the various entity types supported.


gateway-service/src/main/resources/configs/common/application.conf
```diff
@@ -40,39 +40,18 @@ interaction.config = [
 
 timestamp.config = []
 
-domainobject.config = [
+entity.idcolumn.config = [
```
tim-mwangi, line 43: @buchi-busireddy @aaron-steinfeld This is the major transformation

gateway-service/src/main/java/org/hypertrace/gateway/service/GatewayServiceStarter.java
```diff
@@ -34,7 +33,7 @@ protected void doInit() {
             ? getAppConfig().getInt(SERVICE_PORT_CONFIG)
             : DEFAULT_PORT;
 
-    DomainObjectConfigs.init(getAppConfig());
+    //DomainObjectConfigs.init(getAppConfig());
```
tim-mwangi, line 36: We've cut this out

tim-mwangi:
I've tested this rough draft with saas-dev ui.

gateway-service/src/main/resources/configs/common/application.conf
aaron-steinfeld: So we still need the id columns? Doesn't shock me since we do in graphql too, but what's it used for? And can we make a change similar to graphql where we look up id columns from the entity type service and merge them into the hardcoded ones (since the id attribute is part of the metdata for entities that have been migrated to the v2 type service)

tim-mwangi: The entity id is used for the execution tree in order to group by entity id for the aggregation requests. If we want to use entity type service, does that work with dynamic entity types like DOMAIN? Also introducing a new service dependency should be done in a new PR.

aaron-steinfeld: Yep, agree it's a separate piece of work, and it doesn't replace the hardcoded config until all entities are migrated. Entity type service doesn't distinguish between projections, that's orthogonal.
Graphql, implements it as a dynamic loader: https://github.com/hypertrace/hypertrace-graphql/blob/main/hypertrace-graphql-entity-type/src/main/java/org/hypertrace/graphql/entity/EntityIdMappingLoader.java

Which gets combined with the static mappings: https://github.com/hypertrace/hypertrace-core-graphql/blob/3fc65784f6a37bcca8c069eb4e88eee9b063c772/hypertrace-core-graphql-attribute-store/src/main/java/org/hypertrace/core/graphql/attributes/IdLookup.java

tim-mwangi: We used to have this in code as well till we changed to config so that we could support different entities dynamically. So do we proceed with this for now @aaron-steinfeld @buchi-busireddy ?

aaron-steinfeld: sure

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/config/EntityIdColumnsConfigs.java
```diff
@@ -0,0 +1,51 @@
...
+  private static final String DOMAIN_OBJECT_CONFIG = "domainobject.config";
+  private static final String SCOPE = "scope";
+  private static final String KEY = "key";
+
+  private final Map<String, String> scopeToKeyMap;
+
+  public static EntityIdColumnsConfigs fromConfig(Config config) {
+    if (config.hasPath(ENTITY_ID_COLUMN_CONFIG)) {
+      return fromConfigList(config.getConfigList(ENTITY_ID_COLUMN_CONFIG));
+    } else if (config.hasPath(DOMAIN_OBJECT_CONFIG)) {
```
aaron-steinfeld, line 21: so backwards compatible? that seems good.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/config/EntityIdColumnsConfigs.java
```diff
@@ -0,0 +1,51 @@
...
+      return fromConfigList(config.getConfigList(DOMAIN_OBJECT_CONFIG));
+    } else {
+      return new EntityIdColumnsConfigs(Map.of());
+    }
+  }
+
+  private static EntityIdColumnsConfigs fromConfigList(List<? extends Config> configs) {
+    Map<String, String> scopeToKeyMap = new HashMap<>();
+    EntityIdColumnsConfigs entityIdColumnsConfigs = new EntityIdColumnsConfigs(scopeToKeyMap);
+    for (Config scopeConfig : configs) {
```
aaron-steinfeld, line 31: Don't suppose I could convince you to make this immutable? Or at least on line 41 to keep its internal state immutable.

aaron-steinfeld (commented):
Not familiar enough with gateway to approve, but lgtm

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/config/EntityIdColumnsConfigs.java
tim-mwangi: Yeah. I will prepare a helm values PR also but this should go away as well once the helm values PR has gone through the pipeline.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/config/EntityIdColumnsConfigs.java
tim-mwangi: Will do.

buchi-busireddy (approved):

gateway-service/src/main/resources/configs/common/application.conf
buchi-busireddy: Good discussion and my first point as well. Logged https://traceableai.atlassian.net/browse/ENG-7183 so that we don't miss it.

PR #52, opened 2020-12-04
fix: added missing api_trace id mapping

skjindal93:
## Description
The scope filters are added per scope, based on the domainobject configs defined here https://github.com/hypertrace/gateway-service/blob/main/gateway-service/src/main/resources/configs/common/application.conf#L43

The scope filter added for `API_TRACE` being https://github.com/hypertrace/gateway-service/blob/main/gateway-service/src/main/resources/configs/common/application.conf#L81, were not getting applied, since there were no attribute id mappings defined for the `API_TRACE` scope and hence, the API traces returned did not have the `apiBoundaryType = ENTRY` applied on them

### Testing
Tested end to end manually

### Checklist:

### DIFF
--- a/gateway-service/src/main/resources/configs/common/application.conf
+++ b/gateway-service/src/main/resources/configs/common/application.conf
@@ -64,6 +64,17 @@ domainobject.config = [
     ]
   },
   {
+    scope = API_TRACE
+    key = apiTraceId
+    primaryKey = true
+    mapping = [
+      {
+        scope = API_TRACE
+        key = apiTraceId
+      }
+    ]
+  },
+  {
     scope = BACKEND
     key = id
     primaryKey = true


avinashkolluru (approved):

kotharironak:
@skjindal93 Does this impact any functionality in hypertrace side?

skjindal93:
> @skjindal93 Does this impact any functionality in hypertrace side?

@kotharironak Does not impact any functionality, but fixes a functionality. The issue being the API traces returned for an API were returning all the spans, even the internal spans, and, hence there were many duplicate spans and empty URLs

The API traces should only return API traces -> spans with `apiBoundaryType = ENTRY`

The filter was already present to be added based on the `API_TRACE` scope here https://github.com/hypertrace/gateway-service/blob/main/gateway-service/src/main/resources/configs/common/application.conf#L90

```
scopeFiltersConfig = [
  {
    scope = API_TRACE
    filters = [
      {
        scope = API_TRACE
        key = apiBoundaryType
        op = EQ
        value = ENTRY
      },
      {
        scope = API_TRACE
        key = apiId
        op = NEQ
        value = "null"
      }
    ]
  }
]
```

But, the scope filters were not getting applied since the id mappings were missing. Adding the missing id mappings will now apply the scope filters correctly and API traces will return the expected spans

tim-mwangi:
Huh. That's weird. Why do scopeFiltersConfig depend on the entity id configs?

PR #53, opened 2020-12-16
feat(entities): support for time agnostic apis

skjindal93:
## Description
See #54 for more details.

### Testing
Tested end to end locally all the way from GQL by running the service locally

### Checklist:

### DIFF
--- a/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
+++ b/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
@@ -44,7 +44,7 @@ message EntitiesRequest {
 
   int32 limit = 20;
   int32 offset = 21;
-  bool query_live_entities_only = 22;
+  bool include_results_outside_time_range = 22;
 }
 
 message EntitiesResponse {
--- a/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
+++ b/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
@@ -44,7 +44,7 @@ message EntitiesRequest {
 
   int32 limit = 20;
   int32 offset = 21;
-  bool include_results_outside_time_range = 22;
+  bool include_non_live_entities = 22;
 }
 
 message EntitiesResponse {
--- a/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
+++ b/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
@@ -44,6 +44,7 @@ message EntitiesRequest {
 
   int32 limit = 20;
   int32 offset = 21;
+  bool query_live_entities_only = 22;
 }
 
 message EntitiesResponse {
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
@@ -43,20 +43,10 @@ public class RequestPreProcessor {
   public EntitiesRequest transformFilter(
       EntitiesRequest originalRequest, EntitiesRequestContext context) {
     EntitiesRequest.Builder entitiesRequestBuilder = EntitiesRequest.newBuilder(originalRequest);
-    // Convert the time range into a filter and set it on the request so that all downstream
-    // components needn't treat it specially.
-    Filter filter = originalRequest.getFilter();
-    if (originalRequest.getQueryLiveEntitiesOnly()) {
-      filter =
-          TimeRangeFilterUtil.addTimeRangeFilter(
-              context.getTimestampAttributeId(), filter,
-              originalRequest.getStartTimeMillis(), originalRequest.getEndTimeMillis());
-    }
-
     // Add any additional filters that maybe defined in the scope filters config
-    filter = scopeFilterConfigs.createScopeFilter(
+    Filter filter = scopeFilterConfigs.createScopeFilter(
         originalRequest.getEntityType(),
-        filter,
+        originalRequest.getFilter(),
         attributeMetadataProvider,
         context);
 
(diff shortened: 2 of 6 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -53,6 +54,18 @@ public QueryNode build() {
     // All expressions' attributes from the same source. Will only need one downstream query.
     Optional<String> singleSourceForAllAttributes = ExecutionTreeUtils.getSingleSourceForAllAttributes(executionContext);
 
+    EntitiesRequest entitiesRequest = executionContext.getEntitiesRequest();
+
+    // TODO: If there is a filter on a data source, other than EDS, then the flag is a no-op
+    if (!entitiesRequest.getQueryLiveEntitiesOnly()) {
+      QueryNode rootNode =
+          new DataFetcherNode(EDS.name(), entitiesRequest.getFilter());
+      executionContext.removePendingSelectionSource(EDS.name());
```
avinashkolluru, line 63: rootNode.acceptVisitor(new ExecutionContextBuilderVisitor(executionContext));
 instead of the explicit removePending* calls

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -45,9 +45,13 @@ public EntitiesRequest transformFilter(
     EntitiesRequest.Builder entitiesRequestBuilder = EntitiesRequest.newBuilder(originalRequest);
     // Convert the time range into a filter and set it on the request so that all downstream
     // components needn't treat it specially.
-    Filter filter = TimeRangeFilterUtil.addTimeRangeFilter(
-        context.getTimestampAttributeId(), originalRequest.getFilter(),
-        originalRequest.getStartTimeMillis(), originalRequest.getEndTimeMillis());
+    Filter filter = originalRequest.getFilter();
+    if (originalRequest.getQueryLiveEntitiesOnly()) {
```
avinashkolluru, line 49: I'm assuming this check alone is not enough. The TODO you added in ExecutionTreeBuilder is relevant here too right?

skjindal93: right. I was thinking to get rid of the check and let it add the time range filter, but then remove the time range filter specifically in the `ExecutionVisitor` based on the flag

So, that we have the check at a single place

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
```diff
@@ -44,6 +44,7 @@ message EntitiesRequest {
 
   int32 limit = 20;
   int32 offset = 21;
+  bool query_live_entities_only = 22;
```
skjindal93, line 47: The flag is not backward compatible. Because, it would be `false` by default, and we want to query live entities only by default to make it backward compatible.

Looking for better names for the flag, so that it can be reverted and can stay in the backward compatible mode

kotharironak:
Could you please tag a corresponding issue? It will be helpful in release notes and highlighting new features in HT

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -195,6 +207,25 @@ QueryNode buildExecutionTree(ExecutionContext executionContext, QueryNode filter
...
+    // Convert the time range into a filter and set it on the request so that all downstream
+    // components needn't treat it specially
+    Filter timeRangeFilter =
+        TimeRangeFilterUtil.addTimeRangeFilter(
+            context.getTimestampAttributeId(),
+            filter,
+            context.getEntitiesRequest().getStartTimeMillis(),
+            context.getEntitiesRequest().getEndTimeMillis());
+
+    return buildFilterTree(
```
avinashkolluru, line 221: Doesn't `TimeRangeFilterUtil.addTimeRangeFilter` already AND the input filter?
Do we need to explicitly create another filter?

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
```diff
@@ -44,6 +44,7 @@ message EntitiesRequest {
 
   int32 limit = 20;
   int32 offset = 21;
+  bool include_results_outside_time_range = 22;
```
avinashkolluru, line 47: We definitely need some naming experts here :) 
This doesn't seem to register that well but at the same time cant really suggest a better name either.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: Oh right

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
skjindal93: How about `include_non_live_entities`?

skjindal93:
> Could you please tag a corresponding issue? It will be helpful in release notes and highlighting new features in HT

https://github.com/hypertrace/gateway-service/issues/54

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
avinashkolluru: Yes. That sounds a lot better and clear.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -53,6 +55,16 @@ public QueryNode build() {
     // All expressions' attributes from the same source. Will only need one downstream query.
     Optional<String> singleSourceForAllAttributes = ExecutionTreeUtils.getSingleSourceForAllAttributes(executionContext);
 
+    EntitiesRequest entitiesRequest = executionContext.getEntitiesRequest();
+
+    // TODO: If there is a filter on a data source, other than EDS, then the flag is a no-op
+    if (entitiesRequest.getIncludeResultsOutsideTimeRange()) {
```
avinashkolluru, line 61: Please add some comments explaining this special casing.
Even for the below if condition too.

skjindal93: done

avinashkolluru (approved):

PR #55, opened 2020-12-18
Add baseline api and implementaion for entities

pavan-traceable:
## Description
Added API's for baseline values for Entities and v1 implementation for baseline. 

### Testing
Added unit tests for BaselineService. 

### Checklist:

### Documentation


gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,33 @@
...
+option java_package = "org.hypertrace.gateway.service.v1.baseline";
+
+package org.hypertrace.gateway.service.v1.health;
+
+import "org/hypertrace/gateway/service/v1/gateway_query.proto";
+
+message BaselineEntity {
+  string id = 1;
+  string entity_type = 2;
+  map<string, org.hypertrace.gateway.service.v1.common.Value> attribute = 4;
```
aaron-steinfeld, line 13: Attributes are not needed on a baseline entity

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,33 @@
...
+  map<string, org.hypertrace.gateway.service.v1.common.BaselineMetricSeries> baseline_metric_series = 6;
+}
+
+message BaselineEntitiesRequest {
+  string entity_type = 1;
+  sfixed64 start_time_millis = 2;
+  sfixed64 end_time_millis = 3;
+
+  // Filters using which the entities need to be filtered.
+  org.hypertrace.gateway.service.v1.common.Filter filter = 4;
```
aaron-steinfeld, line 24: Last week, we discussed the problem with this approach - did something change?
To resummarize:
If we want to request baselines in parallel with the standard entity requests, we need to provide all of the inputs to fully specify a result set (sorts, limits, etc.) - given that we really don't want to duplicate that type of query processing (especially with the other changes around entity querying in flight - for example time agnostic queries, which @skjindal93 already landed) the option that likely makes more sense is to take in an explicit list of entity ids instead of full filters, and wait to query until after the first entity request comes back.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
```diff
@@ -10,6 +10,7 @@ import "org/hypertrace/gateway/service/v1/gateway_query.proto";
 
 // Entity + metric data
 message Entity {
+  string id = 1;
```
aaron-steinfeld, line 13: wonder why we started with 2...

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
```diff
@@ -172,6 +178,11 @@ message MetricSeries {
   Period period = 3;
 }
 
+message BaselineMetricSeries {
+  Baseline baseline = 2;
```
aaron-steinfeld, line 182: A metric series doesn't have a baseline outside of the intervals - it's fine to approach it iteratively, but let's leave it empty rather than add a placeholder that wouldn't be used.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
```diff
@@ -107,8 +107,7 @@ message Expression {
 }
 
 message HealthExpression {
-  // Currently this is empty since there is nothing additional to provide.
-  // Presence of this expression itself indicates that the health of metric is required.
+  FunctionExpression function = 1;
```
aaron-steinfeld, line 110: What is this? Because the baseline api is separate, we no longer need a message like this (i.e. it can remain defunct)

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
```diff
@@ -155,10 +154,17 @@ message Period {
...
+  Value value = 1;
+  Value lowerBound = 2;
+  Value upperBound = 3;
+}
+
 message AggregatedMetricValue {
   FunctionType function = 1;
   Value value = 2;
   Health health = 3;
+  Baseline baseline = 4;
```
aaron-steinfeld, line 167: This shouldn't change, the baseline and main entity APIs don't overlap by design. Would actually recommend keeping all changes inside the baseline.proto file to make that clear.

aaron-steinfeld (commented):
Reviewed the api only - let's get that settled first.

aaron-steinfeld (changes requested):
Have major concerns on the api - let's take it from there and come back to the impl.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: I thought those changes will take time to finish, so i started with above approach. In the implementation i have just modified the request and call Entity Service, we can easily add sort, limits etc since implementation is same.  But more cleaner approach is the one you suggested, i agree with that. Will the same approach work for attributes as well ?

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
pavan-traceable: Ok. Since health expression is already present i used, if function present inside health expression, then that function is eligible for baseline calculation.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
pavan-traceable: I agree, i added this because merging of EntityResponse and BaslineResponse will be easier in graphql later.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
pavan-traceable: Currently metric series has only one value later it will be added to interval specific.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
aaron-steinfeld: I'm not sure what you mean - given the plan is to move baseline into an entirely separate service, the only scalable way for us to implement this is calling the entity API as is, then turning around with the returned result IDs in GQL and making a request for the baseline values for those IDs (which allows all paging/sorting/filtering to only be implemented in one spot). The attributes (as well as the paging/sorting/filtering) would stay on the first request, as they currently are, and the ID can then be used to correlate the two results and stitch them together.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
aaron-steinfeld: Even though there will only be one interval, returning it as an interval allows us to evolve the implementation without changing the API (or changing the GQL code at all).

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
aaron-steinfeld: Isn't that function already represented by the parent aggregation? (moot point though, given the discussion about separating the APIs)

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
pavan-traceable: @buchi-busireddy  told that initially we have added id and removed since that is also an attribute. It is good to have but i can still get entityId from attributesList.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,52 @@
...
+
+import "org/hypertrace/gateway/service/v1/gateway_query.proto";
+
+message BaselineEntitiesRequest {
+  string entity_type = 1;
+  sfixed64 start_time_millis = 2;
+  sfixed64 end_time_millis = 3;
+
+  repeated string entity_ids = 4;
+  repeated org.hypertrace.gateway.service.v1.common.FunctionExpression baseline_aggregate = 7;
```
aaron-steinfeld, line 16: could we make the naming of these symmetric with the response - e.g. metric_request/metric, metric_series_request/metric_series or something similar (same for the types themselves)

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,52 @@
...
+}
+
+message Baseline {
+  org.hypertrace.gateway.service.v1.common.Value value = 1;
+  org.hypertrace.gateway.service.v1.common.Value lowerBound = 2;
+  org.hypertrace.gateway.service.v1.common.Value upperBound = 3;
+}
+
+message BaselineMetricSeries {
+  repeated BaselineInterval baselineValue = 1;
```
aaron-steinfeld, line 43: this can just return the intervals for now, don't need the other data - it's unused.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
aaron-steinfeld: Gotcha. Since it's special in both graphql and gateway (i.e. we need to know what its value is and the associated attribute key) we don't gain anything by treating it as a plain attribute and instead need to jump through hoops to extract it. So I think it's a good change, but given that the code to work with the generic id is already in place, we may want to keep it separate from this to keep the change footprint down unless it helps in the implementation - the important part is the id on the baseline response.

aaron-steinfeld (commented):
Small comments, but overall api looks good.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: Done.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: In the next iteration instead of one baseline, we would like to generate multiple data points for that period is required.

aaron-steinfeld: This information is on the request, but here we're just echoing back the request (`BaselineTimeAggregation`). The specific period of an interval is already on each as start/end times.

pavan-traceable: ok. Pls review implementation changes as well. I have finished it.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  private long getPeriodInSecs(Period period) {
+    ChronoUnit unit = ChronoUnit.valueOf(period.getUnit());
+    long periodSecs = Duration.of(period.getValue(), unit).getSeconds();
+    return periodSecs;
+  }
+
+  /**
+   * If user selected time range is less than 24 hrs then choose last 24hrs of data for baseline if
```
aaron-steinfeld, line 352: would recommend max of (24h, 2xTR) - to keep things continuous (i.e. don't jump from 24h of data to 48h exactly at the 1d mark.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+   * it is gt 24hours then choose 2X time range for baseline calculation.
+   *
+   * @param startTimeMillis
+   * @param endTimeMillis
+   * @return
+   */
+  private long getUpdatedStartTimeForAggregations(long startTimeMillis, long endTimeMillis) {
+    if ((endTimeMillis - startTimeMillis) < DAY_IN_MILLIS) {
+      startTimeMillis = endTimeMillis - DAY_IN_MILLIS;
```
aaron-steinfeld, line 361: shift should be based on whole range (i.e. for TR [t_0 -> t_1], we want baseline data of [t_0-24h, t_0]. This implementation gives us [t_1-24h, t_1] for baseline data.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+    } else {
+      startTimeMillis = endTimeMillis - (2 * (endTimeMillis - startTimeMillis));
+    }
+    return startTimeMillis;
+  }
+
+  /**
+   * If user selected time range < 6 hours then 3X time range if it is > 6 hours the 2X time range.
```
aaron-steinfeld, line 369: Same general comments as above - keep the data sets continuous.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -130,9 +144,7 @@ public BaselineEntitiesResponse parseQueryResponse(
         new LinkedHashMap<>();
     while (resultSetChunkIterator.hasNext()) {
       ResultSetChunk chunk = resultSetChunkIterator.next();
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Received chunk: " + chunk.toString());
-      }
+      LOG.debug("Received chunk: {} ", chunk.toString());
```
aaron-steinfeld, line 147:
```suggestion
      LOG.debug("Received chunk: {} ", chunk);
```
This removes the need for the if because the expensive stringification only happens if the message is needed to be logged.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      if (chunk.getRowCount() < 1) {
+        break;
+      }
+
+      for (Row row : chunk.getRowList()) {
+        EntityKey entityKey =
+            EntityKey.of(
+                IntStream.range(0, idColumnsSize)
```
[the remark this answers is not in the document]
pavan-traceable: This is the assumption taken across all the places in gateway service, this is copy pasted code from QueryServiceEntityFetcher. We should definitely refactor it and do it across all the places. Will raise a jira for that, but don't want to add to this since it is already big change.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,229 @@
...
+                    .mapToObj(value -> row.getColumn(value).getString())
+                    .toArray(String[]::new));
+
+        Map<String, BaselineMetricSeries.Builder> metricSeriesMap =
+            entityMetricSeriesMap.computeIfAbsent(entityKey, k -> new LinkedHashMap<>());
+
+        BaselineInterval.Builder intervalBuilder = BaselineInterval.newBuilder();
+
+        // Second column is the time column
```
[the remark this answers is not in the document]
pavan-traceable: Replied above

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,229 @@
...
+        Value value =
+            QueryAndGatewayDtoConverter.convertQueryValueToGatewayValue(
+                row.getColumn(idColumnsSize));
+        if (value.getValueType() == ValueType.STRING) {
+          long time = Long.parseLong(value.getString());
+          intervalBuilder.setStartTimeMillis(time);
+          intervalBuilder.setEndTimeMillis(time + TimeUnit.SECONDS.toMillis(periodSecs));
+
+          for (int i = idColumnsSize + 1;
+              i < chunk.getResultSetMetadata().getColumnMetadataCount();
```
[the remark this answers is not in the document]
pavan-traceable: Nope, it is available for all chunks

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
```diff
@@ -1,33 +1,54 @@
...
                 Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(upperBound).build())
             .setValue(
-                Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(meanValue).build())
+                Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(medianValue).build())
             .build();
     return baseline;
   }
+
+  private static double[] getValuesInDouble(List<Value> metricValues) {
+    Preconditions.checkArgument(metricValues.size() > 0,
```
aaron-steinfeld, line 42: Does this mean we return an error response if I ask for the baseline and there's no historical data? That doesn't match how we treat empty data elsewhere.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -93,58 +95,44 @@ public QueryRequest getQueryRequest(
...
     filterBuilder.addAllChildFilter(
         entityIdAttributes.stream()
             .map(
                 entityIdAttribute ->
-                    QueryRequestUtil.createValueInFilter(entityIdAttribute, entityIds))
-            .map(Filter.Builder::build)
+                    QueryRequestUtil.createFilter(
+                        entityIdAttribute,
+                        Operator.IN,
+                        createStringArrayLiteralExpression(new ArrayList<>(entityIds))))
```
aaron-steinfeld, line 111:
```suggestion
                        createStringArrayLiteralExpression(entityIds)))
```
Is there a reason to rewrap the list?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParserTest.java
```diff
@@ -75,6 +75,8 @@ public void testQueryResponse() {
                   {"entity-id-1", "1608525840000", "16.0"},
                   {"entity-id-1", "1608525540000", "17.0"}
                 }));
+    long endTimeInMillis = System.currentTimeMillis();
```
aaron-steinfeld, line 78: Generally a red flag to use a variable value like this in a test. It looks like its only being given as an input (which is why it's not an issue), but that indicates we're not verifying how it's being used.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -183,13 +171,22 @@ public BaselineEntitiesResponse parseQueryResponse(
...
 
-            Value convertedValue =
-                QueryAndGatewayDtoConverter.convertToGatewayValueForMetricValue(
-                    MetricAggregationFunctionUtil.getValueTypeFromFunction(
-                        timeAggregation.getAggregation(), attributeMetadataMap),
-                    attributeMetadataMap,
-                    metadata,
-                    row.getColumn(i));
+            Value convertedValue;
+            // AVGRATE is adding a specific implementation because Pinot does not directly support this function,
```
aaron-steinfeld, line 175: I'm guessing this is duplicated from elsewhere - can it be shared?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
aaron-steinfeld: It's not:
https://github.com/hypertrace/query-service/blob/main/query-service-api/src/main/proto/response.proto#L25-L26 (implementation confirms this comment)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: Because of extra brackets used intellij suggested change, fixed it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
pavan-traceable: Fixed it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+        mergeEntities(baselineEntityAggregatedMetricsMap, baselineEntityTimeSeriesMap);
+    return mergedResponse;
+  }
+
+  private void updateAliasMap(
+      BaselineRequestContext requestContext, List<TimeAggregation> timeAggregations) {
+    timeAggregations.forEach(
+        (timeAggregation ->
+            requestContext.mapAliasToTimeAggregation(
+                timeAggregation.getAggregation().getFunction().getAlias(),
```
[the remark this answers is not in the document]
aaron-steinfeld: Assuming a client structures their request in a certain way is problematic and certainly a bug - even if it exists elsewhere, we should not replicate it in new code and instead file a bug for or fix other locations. Alias is purely a convenience mechanism for the client.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+      period = Period.newBuilder().setUnit("MINUTES").setValue((int) timePeriod).build();
+    } else {
+      long timePeriod = (endTimeMillis - startTimeInMillis) / (1000 * 60 * 60 * 24);
+      period = Period.newBuilder().setUnit("HOURS").setValue((int) timePeriod).build();
+    }
+    return period;
+  }
+
+  private Period getTimeSeriesPeriod(List<BaselineTimeAggregation> baselineTimeSeriesList) {
+    return baselineTimeSeriesList.get(0).getPeriod();
```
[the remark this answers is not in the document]
aaron-steinfeld: Responded elsewhere on this - we need to handle the no data case without erroring.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
aaron-steinfeld: more data is always better, so unless there's performance impacts I'd go for longer. Using only 30 minutes of data to calculate a baseline seems like it defeats the purpose - any behavior change will need to be caught only if it's just beginning (e.g. if my latency spiked a half hour ago, that's already going to dominate my baseline)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+            Baseline baseline = BaselineCalculator.getBaseline(metricValueArray);
+            baselineMap.put(key, baseline);
+          });
+      // Update intervals
+      metricSeriesMap.forEach(
+          (key, value) -> {
+            List<BaselineInterval> intervalList = value.getBaselineValueList();
+            List<BaselineInterval> revisedList = new ArrayList<>();
+            for (BaselineInterval baselineInterval : intervalList) {
+              if (baselineInterval.getStartTimeMillis() >= startTimeMillis) {
```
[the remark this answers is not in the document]
aaron-steinfeld: But aren't baselines based on a time range that extends further into the past (never the future)? So a calculated interval should never be newer than the requested.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParserTest.java
pavan-traceable: Yes, we are not verifying startTime and endTime in test cases. Added basic tests, more tests can be added in next iteration.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
aaron-steinfeld: Not sure what you mean - if I ask for, say, 23h of data - the baseline will be calculated using 24h of data (and with my other comment - it actually means it will be calculated with only one hour of historical data, since we're calculate the baseline time range incorrectly based on the time range end). On the other hand, asking for 24h of data, the baseline is calculated using 48h (but again, actually only t_0-24h)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
aaron-steinfeld: I think this one is important to fix before merging (see my comment above)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
aaron-steinfeld: ok

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
aaron-steinfeld: The primary comment here is that the comment doesn't match the logic

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: I agree your point, but in UI you cannot select 23 hours, either you can select 12 hours or 24 hours. But anyway this is small change i will do it

aaron-steinfeld: Those are the shortcut relative time ranges - the UI also supports any custom time range

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
```diff
@@ -229,6 +237,55 @@ public void updateEntity(
...
+  public void getBaselineForEntities(
+      BaselineEntitiesRequest request, StreamObserver<BaselineEntitiesResponse> responseObserver) {
+    Optional<String> tenantId =
+        org.hypertrace.core.grpcutils.context.RequestContext.CURRENT.get().getTenantId();
+    if (tenantId.isEmpty()) {
+      responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
+      return;
+    }
+
+    try {
```
tim-mwangi, line 250: Maybe move validations to a class or method of their own like https://github.com/hypertrace/gateway-service/blob/main/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/validators/request/RequestValidator.java

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  private static Baseline getBaseline(double[] metricValueArray) {
+    Median median = new Median();
+    StandardDeviation standardDeviation = new StandardDeviation();
+    double medianValue = median.evaluate(metricValueArray);
+    double sd = standardDeviation.evaluate(metricValueArray);
+    double lowerBound = medianValue - (2 * sd);
+    if (lowerBound < 0) {
+      lowerBound = 0;
```
tim-mwangi, line 29: Not covered by unit tests.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+                Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(medianValue).build())
+            .build();
+    return baseline;
+  }
+
+  private static double[] getValuesInDouble(List<Value> metricValues) {
+    ValueType valueType = metricValues.get(0).getValueType();
+    switch (valueType) {
+      case LONG:
+        return metricValues.stream().mapToDouble(value -> (double) value.getLong()).toArray();
```
tim-mwangi, line 48: Not covered by unit tests

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+      case DOUBLE:
+        return metricValues.stream().mapToDouble(value -> value.getDouble()).toArray();
+      default:
+        throw new IllegalArgumentException("Unsupported valueType " + valueType);
```
tim-mwangi, line 52: Not covered by unit tests

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+            .setUpperBound(
+                Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(upperBound).build())
+            .setValue(
+                Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(medianValue).build())
+            .build();
+    return baseline;
+  }
+
+  private static double[] getValuesInDouble(List<Value> metricValues) {
+    ValueType valueType = metricValues.get(0).getValueType();
```
tim-mwangi, line 45: So, this cannot be a mix of longs and doubles, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,360 @@
...
+
+  /**
+   * If user selected time range is less than 24 hrs then choose last 24hrs of data for baseline if
+   * it is gt 24hours then choose 2X time range for baseline calculation.
+   */
+  private long getUpdatedStartTimeForAggregations(long startTimeMillis, long endTimeMillis) {
+    if ((endTimeMillis - startTimeMillis) < DAY_IN_MILLIS) {
+      startTimeMillis = endTimeMillis - DAY_IN_MILLIS;
+    } else {
+      startTimeMillis = endTimeMillis - (2 * (endTimeMillis - startTimeMillis));
```
tim-mwangi, line 351: Unit test

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
pavan-traceable: No this cannot be mix of double and longs, long for SUM function and double for others.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
pavan-traceable: I have verified locally it is available for all chunks and same code is running in prod as well.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
pavan-traceable: Added unit test

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
pavan-traceable: Added unit test

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
pavan-traceable: Added Unit test

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: Fixed this

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: Fixed this.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParserTest.java
pavan-traceable: StartTime and endTime verified separately in testcases.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: added unit test.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: Since we used now Max(24h, 2XtimeRange) even we are using [t_1-24h, t_1] it will give decent data for baseline and also if we don't include current data we might increase in traffic scenarios. We can also check how this is going and iterate on this.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
pavan-traceable: Fixed and added unitests

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParserTest.java
pavan-traceable: Updated this to use startTimes like other tests.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: Created https://traceableai.atlassian.net/browse/ENG-7467

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: Suppose user selected last 1 hour we calculate time series for last 24 hours, here we are having this condition to send baseline for last 1 hour data instead of sending all 24 hours.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: It is handled properly, if (originalRequest.getBaselineMetricSeriesRequestCount() > 0) in calling method. In that case empty response will be returned.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
pavan-traceable: Removed the comment to avoid confusion.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
pavan-traceable: Same logic exists everywhere and its already tested and it is working as expected. Not sure why it is like that. It will be fixed as part of https://traceableai.atlassian.net/browse/ENG-7467

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineCalculator.java
tim-mwangi: ok

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,219 @@
...
+                      .collect(
+                          Collectors.toMap(
+                              Map.Entry::getKey, e -> getSortedMetricSeries(e.getValue()))));
+      baselineEntities.add(entityBuilder.build());
+    }
+    return BaselineEntitiesResponse.newBuilder().addAllBaselineEntity(baselineEntities).build();
+  }
+
+
+  BaselineMetricSeries getSortedMetricSeries(BaselineMetricSeries.Builder builder) {
```
tim-mwangi, line 214: make private

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,219 @@
...
+            BaselineTimeAggregation timeAggregation =
+                requestContext.getTimeAggregationByAlias(metadata.getColumnName());
+
+            if (timeAggregation == null) {
+              LOG.warn("Couldn't find an aggregate for column: {}", metadata.getColumnName());
+              continue;
+            }
+
+            Value convertedValue;
+            // AVG_RATE is adding a specific implementation because Pinot does not directly support this function,
```
tim-mwangi, line 176: Move comment to where the method is implemented

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
```diff
@@ -0,0 +1,219 @@
...
+            ColumnMetadata metadata = chunk.getResultSetMetadata().getColumnMetadata(i);
+            BaselineTimeAggregation timeAggregation =
+                requestContext.getTimeAggregationByAlias(metadata.getColumnName());
+
+            if (timeAggregation == null) {
+              LOG.warn("Couldn't find an aggregate for column: {}", metadata.getColumnName());
+              continue;
+            }
+
+            Value convertedValue;
```
tim-mwangi, line 175: Combine lines 175 and 178

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,362 @@
...
+        AttributeMetadataUtil.getTimestampAttributeId(
+            attributeMetadataProvider, requestContext, originalRequest.getEntityType());
+    Map<String, BaselineEntity> baselineEntityAggregatedMetricsMap = new HashMap<>();
+    Map<String, BaselineEntity> baselineEntityTimeSeriesMap = new HashMap<>();
+
+    if (originalRequest.getBaselineAggregateRequestCount() > 0) {
+      // Aggregated Functions data
+      Period aggTimePeriod =
+          getPeriod(originalRequest.getStartTimeMillis(), originalRequest.getEndTimeMillis());
+      long periodSecs = getPeriodInSecs(aggTimePeriod);
```
tim-mwangi, line 80: Why do we need a period for baselines for simple aggregations(non-time series)?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
```diff
@@ -113,4 +116,27 @@ public static AttributeKind getValueTypeFromFunction(
...
+  public static Value getValueFromFunction(
+          long startTime,
+          long endTime,
+          Map<String, AttributeMetadata> attributeMetadataMap,
+          org.hypertrace.core.query.service.api.Value column,
+          ColumnMetadata metadata,
+          FunctionExpression functionExpression) {
+    Value convertedValue;
+    if (FunctionType.AVGRATE == functionExpression.getFunction()) {
+      convertedValue =
```
tim-mwangi, line 129: Unit test

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
```diff
@@ -113,4 +116,27 @@ public static AttributeKind getValueTypeFromFunction(
         return metadata.getValueKind();
     }
   }
+
+  public static Value getValueFromFunction(
```
tim-mwangi, line 120: Also this can replace RequestHandler.getValueForFunctionExpression() in the explore impl

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: Done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
pavan-traceable: done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: Simple aggregations converted into time series data to calculate standard deviation which will be used to calculate lowerBound and upperBound.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
pavan-traceable: This exists already in QueryServiceEntityFetcher which will have unit tests, this is not anything new.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
pavan-traceable: I have searched for RequestHandler.getValueForFunctionExpression() in gateway Service couldn't find anything

tim-mwangi: https://github.com/hypertrace/gateway-service/blob/cf27396552efd5e12383ccd7e1079997beda81e1/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java#L296

pavan-traceable: Sorry my bad. Fixed this can you please check now.

pavan-traceable:
@aaron-steinfeld  @tim-mwangi please approve if there aren't any concerns, i have to update gateway service api version in  graphql changes.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
tim-mwangi: codecov was saying previously that it was not covered.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
tim-mwangi: ok

tim-mwangi:
lgtm. Will let @aaron-steinfeld approve if he's ok with since he has blocked it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
tim-mwangi: @pavan-traceable what happens when query-service returns an empty response?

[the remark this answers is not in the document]
aaron-steinfeld: ok - I traced this back and I believe the confusion to be a consequence of poor variable names - I assumed baselineTimeSeriesList was a list of timeSeries rather than a list of timeSeriesRequests. I think this could benefit from a rename, but not a blocking issue.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
aaron-steinfeld: Is fixing the spread of this into the baseline code feasible?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
aaron-steinfeld: So doesn't that mean this is relying on the incorrect implementation of baseline time range (that is, it overlapping the requested time range)? If I'm reading this right, it's using whatever intervals were requested for the baseline input data (an implementation detail), filtering them, then cloning them as new intervals, stuffing the calculated baseline into each one.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+      List<TimeAggregation> timeAggregations = getTimeAggregationsForAggregateExpr(originalRequest);
+      updateAliasMap(requestContext, timeAggregations);
+      // Take more data to calculate baseline and standard deviation.
+      long seriesStartTime =
```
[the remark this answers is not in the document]
aaron-steinfeld: This is the issue from the last round of comments that I'm most concerned about. It's sourcing the data incorrectly as a baseline shouldn't use data from the time range it's applicable for, and worse - the later implementation implicitly relies on that behavior (see my other comment).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
aaron-steinfeld: I'm confused, because I know this behavior is still around. I reimplemented this myself fairly recently, and left the behavior as it was to maintain compatibility - https://github.com/hypertrace/query-service/blob/main/query-service-impl/src/main/java/org/hypertrace/core/query/service/RowChunkingOperator.java Are you sure you're requesting enough data to actually get multiple chunks?

Anyway, isn't it easy to fix regardless?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
aaron-steinfeld: Responded to this one on one of the other threads - https://github.com/hypertrace/gateway-service/pull/55#discussion_r552241794

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: Suppose user selected last 6 hours of data and the traffic increased from last 4 hours in user environment, if we don't include the current data for baseline won't the values deviate too much from baseline ?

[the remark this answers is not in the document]
aaron-steinfeld: This is precisely the use case of baselines as I understand them, and why I've really been pushing for this change - seeing that the data in the requested range is higher than it historically has been or could be predicted. If the calculation includes the current data, then as you pointed out, it would be more difficult to identify deviations.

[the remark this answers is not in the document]
pavan-traceable: ok, will update this.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: Will fix this, as part of updating endTime.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: updated variable name

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
pavan-traceable: Fixed this.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceQueryParser.java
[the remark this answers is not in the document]
pavan-traceable: Fixed this as well.

avinashkolluru (approved):
Approving to unblock @pavan-traceable

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/baseline/BaselineServiceImpl.java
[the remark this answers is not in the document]
pavan-traceable: We need mapping between column names and function expression either to calculate average_rate value or any other, this is happening through aliases in gateway repo. Even column name is also set from Alias. This is not small change needs more thought process to completely get rid of aliases. I am not sure who introduces aliases and what is the thought process behind this.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,50 @@
...
+  sfixed64 start_time_millis = 2;
+  sfixed64 end_time_millis = 3;
+
+  repeated string entity_ids = 4;
+  repeated org.hypertrace.gateway.service.v1.common.FunctionExpression baseline_aggregate_request = 7;
+  repeated BaselineTimeAggregation baseline_metric_series_request = 8;
+}
+
+message BaselineEntitiesResponse {
+  repeated BaselineEntity baselineEntity = 1;
```
skjindal93, line 21: Shouldn't these be in Snake Case? Like `baseline_entity`?

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,50 @@
...
+}
+
+message BaselineTimeAggregation {
+  org.hypertrace.gateway.service.v1.common.Period period = 1;
+  org.hypertrace.gateway.service.v1.common.FunctionExpression aggregation = 2;
+}
+
+message Baseline {
+  org.hypertrace.gateway.service.v1.common.Value value = 1;
+  org.hypertrace.gateway.service.v1.common.Value lowerBound = 2;
```
skjindal93, line 38: Similarly here, Should be `lower_bound` and `upper_bound` I believe

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
```diff
@@ -0,0 +1,50 @@
...
+  org.hypertrace.gateway.service.v1.common.Value upperBound = 3;
+}
+
+message BaselineMetricSeries {
+  repeated BaselineInterval baselineValue = 1;
```
skjindal93, line 43: `baseline_value`

pavan-traceable: done

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: done

skjindal93 (approved):

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/baseline.proto
pavan-traceable: done

avinashkolluru (approved):

PR #58, opened 2020-12-30
chore: using data fetcher node instead of selection and filter node

skjindal93:
## Description
- Gets rid of the `SelectionAndFilterNode` which was redundant and can be replaced with `DataFetcherNode` by adding `limit`, `offset` and `orderBys`
- Pushes down the pagination (`orderBy`, `limit` and `offset`) to EDS, since Entity Service has started supporting pagination
- Fixes a bug, where limit and offset was being pushed down to QS, even if there is a filter on an attribute in EDS

### Testing
Added Unit Tests. And, tested every UI screen locally by pointing to local Gateway Service

### Checklist:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -120,27 +121,35 @@ private QueryNode buildExecutionTreeForSameSourceFilterAndSelection(String sourc
...
     // Pinot will do the ordering for us.
+    // https://github.com/apache/incubator-pinot/issues/111#issuecomment-214810551
     if (selectionOffset > 0) {
       selectionLimit = selectionOffset + selectionLimit;
       selectionOffset = 0;
     }
 
-    QueryNode rootNode = new SelectionAndFilterNode(source, selectionLimit, selectionOffset);
+    QueryNode rootNode =
+        new DataFetcherNode(source, filter, selectionLimit, selectionOffset, orderBys);
```
avinashkolluru, line 141: ExecutionVisitor for SelectionAndFilterNode was doing multiple things
1. getEntitiesAndAggregratedMetrics
2. getTimeAggregatedMetrics

With this change, I dont see the second call being made at all. I think we should add a SelectionNode after the PaginateOnlyNode to fetch time aggregated metrics too.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -170,6 +191,12 @@ public EntityFetcherResponse visit(SelectionNode selectionNode) {
                   EntitiesRequest.newBuilder(executionContext.getEntitiesRequest())
                       .clearSelection()
                       .clearFilter()
+                      // TODO: Should we push order by, limit and offet down to the data source?
+                      // If we want to push the order by down, we would also have to divide order by into
+                      // sourceToOrderBySelectionExpressionMap, sourceToOrderByMetricExpressionMap, sourceToOrderByTimeAggregationMap
```
avinashkolluru, line 196: sourceToOrderByTimeAggregationMap probably doesn't make sense. Time aggregations is for fetching time series where order by is not relevant.
But, the other two are valid and I think should be done. If not in this PR a separate one can follow.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: That's being taken care of by `return buildExecutionTree(executionContext, rootNode);`

avinashkolluru: Ahhh.. Nice. Missed that.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -78,12 +78,13 @@ public QueryNode build() {
     // the data store
     if (singleSourceForAllAttributes.isPresent()) {
       String source = singleSourceForAllAttributes.get();
-      QueryNode selectionAndFilterNode = buildExecutionTreeForSameSourceFilterAndSelection(source);
+      QueryNode rootNode = buildExecutionTreeForSameSourceFilterAndSelection(source);
       if (LOG.isDebugEnabled()) {
-        LOG.debug("Execution Tree:{}", selectionAndFilterNode.acceptVisitor(new PrintVisitor()));
+        LOG.debug("Execution Tree:{}", rootNode.acceptVisitor(new PrintVisitor()));
```
avinashkolluru, line 83: Not related to this change but can we add a similar log for the above `if (entitiesRequest.getIncludeNonLiveEntities())` condition too

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -78,12 +78,13 @@ public QueryNode build() {
...
       }
 
-      return selectionAndFilterNode;
+      rootNode.acceptVisitor(new ExecutionContextBuilderVisitor(executionContext));
+      return buildExecutionTree(executionContext, rootNode);
```
avinashkolluru, line 87: This is much cleaner now. Can you just add a comment on top of the `filterTree.acceptVisitor(new OptimizingVisitor())` at line 100. Mention that this is needed to merge filters corresponding to the same source into one DataFetcherNode instead of multiple for each filter.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -185,7 +197,12 @@ public EntityFetcherResponse getAggregatedMetrics(
 
     QueryRequest.Builder builder =
         constructSelectionQuery(requestContext, entitiesRequest, entityIdAttributes, aggregates);
-    adjustLimitAndOffset(builder, entitiesRequest.getLimit(), entitiesRequest.getOffset());
+    adjustLimitAndOffset(
```
avinashkolluru, line 200: Not related to this change again but the getEntities and getAggregatedMetrics are very similar. Can we extract  common code out of these methods?

skjindal93: Right. Will try to make it a part of separate PR, so that the regression impact is minimal and separated out

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
```diff
@@ -2,6 +2,10 @@
 
 import org.hypertrace.gateway.service.entity.query.visitor.Visitor;
 import org.hypertrace.gateway.service.v1.common.Filter;
+import org.hypertrace.gateway.service.v1.common.OrderByExpression;
+
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Node in the execution tree that applies the encapsulated filters and fetches attributes
```
avinashkolluru, line 11: Edit the doc to mention it also applies order/limit/offset conditionally if required

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -100,7 +100,19 @@ public EntityFetcherResponse getEntities(
...
+        builder,
+        entitiesRequest.getLimit(),
+        entitiesRequest.getOffset(),
+        requestContext.canApplyLimit(),
+        requestContext.canApplyOffset());
+
+    if (!entitiesRequest.getOrderByList().isEmpty()) {
+      // Order by from the request.
+      builder.addAllOrderBy(
```
avinashkolluru, line 112: Can you add tests for this?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
```diff
@@ -197,6 +198,7 @@ private void mock(AttributeMetadataProvider attributeMetadataProvider) {
                     .build()));
   }
 
+  @Disabled
```
avinashkolluru, line 201: Why is this disabled?

skjindal93: This is Disabled, because currently even if the attributes are to be fetched from the single source, it would still query all the sources, if there is an attribute which lies in multiple sources

The reason being it adds a SelectionNode in the execution tree, since there is a pending selection source.

It needs an additional visitor which would traverse the execution tree and then remove the pending selection sources if the attribute was already requested from a different source

Till then, I decided to keep the test disabled, and enable it back, once we add the new Visitor to prune the pending selection sources from the execution tree

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -78,12 +85,17 @@ public QueryNode build() {
     // the data store
     if (singleSourceForAllAttributes.isPresent()) {
```
skjindal93, line 86: Once we have the optimization of pushing the pagination down if the order by and filter are on the same source, we can actually get rid of the entire single source check

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -120,27 +140,35 @@ private QueryNode buildExecutionTreeForSameSourceFilterAndSelection(String sourc
...
     if (selectionOffset > 0) {
       selectionLimit = selectionOffset + selectionLimit;
       selectionOffset = 0;
     }
 
-    QueryNode rootNode = new SelectionAndFilterNode(source, selectionLimit, selectionOffset);
+    QueryNode rootNode =
+        new DataFetcherNode(source, filter, selectionLimit, selectionOffset, orderBys);
+
     if (executionContext.getEntitiesRequest().getOffset() > 0) {
```
avinashkolluru, line 162: nit: if (selectionOffset > 0)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntitiesRequestContext.java
```diff
@@ -28,6 +31,22 @@ public String getTimestampAttributeId() {
     return timestampAttributeId;
   }
 
+  public boolean canApplyLimit() {
+    return applyLimit;
+  }
+
+  public void canApplyLimit(boolean applyLimit) {
```
avinashkolluru, line 38: nit: these method names are a bit confusing. The fact that one is a get and another is a set is not clear.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: Can't do `if (selectionOffset > 0)`

There is a code block above which says

```
if (selectionOffset > 0) {
  selectionLimit = selectionOffset + selectionLimit;
  selectionOffset = 0;
}
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntitiesRequestContext.java
skjindal93: got rid of the methods. no longer needed as discussed

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
avinashkolluru: Ahh sorry.

avinashkolluru (approved):

PR #60, opened 2021-01-04
optimize filter and order by single source

skjindal93:
## Description
If filters and order bys are requested from the same data source, the pagination can be pushed down to the data store, instead of paginating all the data in memory

### Testing
Added unit tests and verified by running services locally and testing all UI screens with various functionalities

### Checklist:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -232,10 +234,12 @@ private void buildSourceToExpressionMaps() {
       Expression expression = orderByExpression.getExpression();
       Map<String, List<Expression>> map =
           getDataSourceToExpressionMap(Collections.singletonList(expression));
-      // There should only be one element in the map.
```
skjindal93, line 235: The assumption was wrong. The map could have more elements, if an attribute is present in multiple sources. Does not affect anything at all

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
```diff
@@ -179,4 +187,131 @@ private static boolean containsNotOrOrFilter(Filter filter) {
...
+  /**
+   * Given a source to expression map, builds an attribute to sources map, where attribute value is extracted
+   * out from the expression. Basically, a reverse map of the map provided as input
+   *
+   * Example:
+   * ("QS" -> API.id, "QS" -> API.name, "EDS" -> API.id) =>
+   *
+   * ("API.id" -> ["QS", "EDS"], "API.name" -> "QS")
+   */
+  private static Map<String, Set<String>> buildAttributeToSourcesMap(
```
avinashkolluru, line 235: Isn't this map already there in ExecutionContext?
See attributeToSourcesMap in ExecutionContext

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/OptimizingVisitor.java
```diff
@@ -155,7 +155,20 @@ public QueryNode visit(SortAndPaginateNode sortAndPaginateNode) {
                               .addAllChildFilter(filterList)
                               .build();
                     }
-                    return new DataFetcherNode(stringListEntry.getKey(), filter);
+                    // There should always be at least one entry
+                    if (!stringListEntry.getValue().isEmpty()) {
+                      String source = stringListEntry.getKey();
+                      DataFetcherNode dataFetcherNode = (DataFetcherNode) stringListEntry.getValue().get(0);
```
avinashkolluru, line 161: Trying to understand why this special casing is done? What changes led to this?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
skjindal93: Wondering why I missed that. But, yes, I think I can use that to filter out the filter and order by attributes, instead of building the entire map again

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/OptimizingVisitor.java
skjindal93: It was earlier creating a new `DataFetcherNode` which is fine, but since we have now added `limit`, `offset` and `orderBy` in DataFetcherNode, we would need access to `limit`, `offset` and `orderBys`, so that a new DataFetcherNode can be created with those values

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
```diff
@@ -29,8 +29,8 @@ public DataFetcherNode(String source, Filter filter) {
   public DataFetcherNode(
       String source,
       Filter filter,
-      int limit,
-      int offset,
+      Integer limit,
```
tim-mwangi, line 32: why?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
```diff
@@ -186,4 +190,129 @@ private static boolean containsNotOrOrFilter(Filter filter) {
...
+    // merges sourceToSelectionOrderByAttributeMap and sourceToMetricOrderByAttributeMap
+    Map<String, Set<String>> sourceToOrderByAttributeMap =
+        Stream.concat(
+            sourceToSelectionOrderByAttributeMap.entrySet().stream(),
+            sourceToMetricOrderByAttributeMap.entrySet().stream())
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    Map.Entry::getValue,
+                    (v1, v2) -> {
```
tim-mwangi, line 216: Can we make sure that all scenarios in this class are unit tested?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/FilterOptimizingVisitor.java
```diff
@@ -155,7 +155,20 @@ public QueryNode visit(SortAndPaginateNode sortAndPaginateNode) {
...
+                      DataFetcherNode dataFetcherNode = (DataFetcherNode) stringListEntry.getValue().get(0);
+                      return
+                          new DataFetcherNode(
+                              source,
+                              filter,
+                              dataFetcherNode.getLimit(),
+                              dataFetcherNode.getOffset(),
+                              dataFetcherNode.getOrderByExpressionList());
+                    } else {
+                      return new NoOpNode();
```
tim-mwangi, line 170: Test for this negative scenario?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceTest.java
```diff
@@ -196,7 +196,7 @@ private void mock(AttributeMetadataProvider attributeMetadataProvider) {
                     .addSources(AttributeSource.QS)
                     .build()));
   }
-
+
```
tim-mwangi, line 199: Nit: undo

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
skjindal93: I had to make this change, so that we can create a new `DataFetcherNode` in `FilterOptimizingVisitor`

```
return
                          new DataFetcherNode(
                              source,
                              filter,
                              dataFetcherNode.getLimit(),
                              dataFetcherNode.getOffset(),
                              dataFetcherNode.getOrderByExpressionList());
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
skjindal93: yep. I have added unit tests for all the scenarios

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/FilterOptimizingVisitor.java
skjindal93: The negative scenario should never occur, but I can still add a unit test for the case

skjindal93: Can't even add a unit test, since the case will never occur

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -367,6 +378,15 @@ private void getSourceToFilterExpressionMap(Filter filter,
...
 
+  /**
+   * Given a source to expression map, builds an attribute to sources map, where attribute value is extracted
+   * out from the expression. Basically, a reverse map of the map provided as input
+   *
+   * Example:
+   * ("QS" -> API.id, "QS" -> API.name, "EDS" -> API.id) =>
+   *
+   * ("API.id" -> ["QS", "EDS"], "API.name" -> "QS")
+   */
```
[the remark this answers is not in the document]
skjindal93: done

avinashkolluru (approved):

PR #61, opened 2021-01-04
fix(entities): total entites was never fetched for composite sources 

skjindal93:
## Description
Total was never fetched specifically for data coming from multiple data sources. The total was only set in the execution context while visiting some nodes, but those nodes used to lie a level above `PaginationNode`s, hence cutting down the correct total

DataFetcherNode -> SelectionNode -> SortAndPaginateNode -> SelectionNode (_this node would cut down the total to the number of paginated results_)

Hence, the total was always equals to `min(number of results, limit)`

- Total should never be set in SelectionNode or PaginationNode
- Total should always be fetched
- `TotalFetcherNode` was redundant. `DataFetcherNode` itself should ask for total number of results with the same filter, but no offset and limit
- `AndNode` or `OrNode` can change the number of the total results, because the data is coming from multiple data sources
- Setting to total as always 1 if there is an entity id filter is wrong. What if the entity id does not even exist in the system?
- Setting total in execution context is no longer needed

### Testing
- Verified locally by port forwarding various services to serve the UI
- This would need a projection support on document-store, for the total query to be efficient from EDS, otherwise it just fetches all the API entity documents, even though it just needs to fetch only the set of entity keys to compute the total

### Checklist:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -141,42 +172,54 @@ public EntityFetcherResponse visit(DataFetcherNode dataFetcherNode) {
 
     EntitiesRequest request = requestBuilder.build();
     IEntityFetcher entityFetcher = queryHandlerRegistry.getEntityFetcher(source);
-    return entityFetcher.getEntities(context, request);
+
+    EntitiesRequest totalEntitiesRequest =
```
avinashkolluru, line 176: Shouldn't the total be computed conditionally based on whether it is requested or not?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -141,42 +172,54 @@ public EntityFetcherResponse visit(DataFetcherNode dataFetcherNode) {
...
+        EntitiesRequest.newBuilder(executionContext.getEntitiesRequest())
+            .clearSelection()
```
avinashkolluru, line 178: Optimization: If the limit is not passed down to the EntityFetcher, can we avoid this extra query?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -19,7 +19,9 @@
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import com.google.common.collect.Sets;
 import org.hypertrace.gateway.service.common.datafetcher.EntityFetcherResponse;
+import org.hypertrace.gateway.service.common.datafetcher.EntityResponse;
```
avinashkolluru, line 24: Where is this EntityResponse? I dont see it anywhere in the PR

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
avinashkolluru: Ingore. I think we always send the total.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -320,14 +368,13 @@ Filter constructFilterFromChildNodesResult(EntityFetcherResponse result) {
   }
 
   @Override
-  public EntityFetcherResponse visit(SortAndPaginateNode sortAndPaginateNode) {
-    EntityFetcherResponse result = sortAndPaginateNode.getChildNode().acceptVisitor(this);
-    // Set the total entities count in the execution context before pagination
-    executionContext.setTotal(result.size());
```
avinashkolluru, line 326: Isn't this still needed? What if there is no DataFetcherNode at all. Who sets this in such a case?

skjindal93: Good catch. The current scenarios always had a DataFetcherNode, but yes, I will add it here as well. That will account for the optimization here https://github.com/hypertrace/gateway-service/pull/61#discussion_r556570387

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
skjindal93: oops. I forgot to Git add it. Pushed now

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
skjindal93: Yes. But, we can be smart. We can have a boolean flag on the entities request which would be set when GQL asks for the total, else if the GQL query does not ask for `total`, we won't fetch the total.

For cases, like entity id filter, UI doesn't need total at all, so it doesn't even fetch the total via GQL

avinashkolluru: Yeah... lets do that

skjindal93: let me do that as a part of follow up PR, because that would anyways need a change in GQL layer as well

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
```diff
@@ -182,17 +183,8 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
...
-            .clearSelection()
-            .clearTimeAggregation()
-            .clearOrderBy()
-            .clearLimit()
-            .setOffset(0)
-            .build()
-    );
-    return entityFetcherResponse.size();
+  public Set<EntityKey> getTotalEntities(EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
+    EntityFetcherResponse entityFetcherResponse = getEntities(requestContext, entitiesRequest);
```
avinashkolluru, line 187: Wouldn't this fetch the entire entity. Why not just project the entity id?

avinashkolluru: Sorry. Ignore. Just saw that it is taken care of in DataFetcherNode.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -554,28 +555,21 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
   }
 
   @Override
-  public int getTotalEntities(EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
+  public Set<EntityKey> getTotalEntities(EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
```
avinashkolluru, line 558: Is there a point in retaining this method when we are anyways relying on getEntities. Should we change the callers itself to use getEntities instead?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
```diff
@@ -182,17 +183,8 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
...
-        EntitiesRequest.newBuilder(entitiesRequest)
-            .clearSelection()
-            .clearTimeAggregation()
-            .clearOrderBy()
-            .clearLimit()
-            .setOffset(0)
-            .build()
-    );
-    return entityFetcherResponse.size();
+  public Set<EntityKey> getTotalEntities(EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
```
tim-mwangi, line 186: This method implies that a number is the return value. Can it be renamed to reflect that we are returning a set?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
skjindal93: This gives us a flexibility to modify the total entities request. For example: QueryServiceEntityFetcher is setting a limit of 10k on the total entities request.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
```diff
@@ -37,6 +40,7 @@ public DataFetcherNode(
     this.limit = limit;
     this.offset = offset;
     this.orderByExpressionList = orderByExpressionList;
+    this.isPaginated = limit != null && offset != null;
```
tim-mwangi, line 43: Given that the initial source of this is proto API, how do you know when it's not requested?

skjindal93: when the creator of DataFetcherNode does not pass the limit and offset, it wouldn't be paginated

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/EntitiesRequestAndResponseUtils.java
```diff
@@ -142,6 +145,21 @@ public static void compareEntityFetcherResponses(EntityFetcherResponse expectedE
     });
   }
 
+  public static void compareEntityKeys(
+      Set<EntityKey> expectedEntityKeys, Set<EntityKey> actualEntityKeys) {
+    assertEquals(expectedEntityKeys.size(), actualEntityKeys.size());
```
tim-mwangi, line 150: We can't use set.equals()?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
skjindal93: makes sense. done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
skjindal93: in fact, got rid of the method. wasn't needed https://github.com/hypertrace/gateway-service/pull/61#discussion_r558370431

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/EntitiesRequestAndResponseUtils.java
skjindal93: done

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
```diff
@@ -207,22 +208,17 @@ public void test_getTotalEntitiesSingleEntityIdAttribute() {
     String tenantId = "TENANT_ID";
     Map<String, String> requestHeaders = Map.of("x-tenant-id", tenantId);
     AttributeScope entityType = AttributeScope.API;
-    EntitiesRequest entitiesRequest =
```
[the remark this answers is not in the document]
skjindal93: no longer valid, since we got rid of total entities method

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcherTests.java
```diff
@@ -122,6 +122,51 @@ public void test_getEntities_WithPagination() {
         entityDataServiceEntityFetcher.getEntities(entitiesRequestContext, entitiesRequest).size());
   }
 
+  @Test
+  public void test_getTotalEntities() {
```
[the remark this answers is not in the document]
skjindal93: I don't think so. Removed

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
tim-mwangi: ok.

tim-mwangi (approved):

avinashkolluru (approved):

PR #63, opened 2021-01-11
don't fetch non live entities, if there are filters on non EDS data source

skjindal93:
## Description
If there are filters on other data sources other than EDS, do not fetch APIs time agnostically, even if `setNonLiveEntities` is set to true

### Testing
Added unit tests for all cases

### Checklist:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
```diff
@@ -191,24 +206,43 @@ private static boolean containsNotOrOrFilter(Filter filter) {
     return false;
   }
 
+  public static boolean areFiltersOnlyOnCurrentDataSource(ExecutionContext executionContext, String currentSource) {
```
skjindal93, line 209: The only function added. Rest are formatting changes in the class

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
```diff
@@ -191,24 +206,43 @@ private static boolean containsNotOrOrFilter(Filter filter) {
...
+    Map<String, Set<String>> sourceToFilterAttributeMap =
+        executionContext.getSourceToFilterAttributeMap();
+
+    if (sourceToFilterAttributeMap.isEmpty()) {
+      return true;
+    }
+
+    Map<String, Set<String>> filterAttributeToSourcesMap =
+        buildAttributeToSourcesMap(sourceToFilterAttributeMap);
```
avinashkolluru, line 218: I see this map being constructed at other places too. Does it make sense to save this within the ExecutionContext and reuse instead of computing always?

skjindal93: I can do it for the filter, since that's constructed twice everytime

skjindal93: done

avinashkolluru (approved):

PR #67, opened 2021-01-17
fix(entities): incorrect source to filter expression map

skjindal93:
## Description
Filter maps in Execution Context was being built incorrectly

### Testing
Added unit tests and verified end to end

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
@@ -323,8 +323,8 @@ public class ExecutionContext {
     return ImmutableMap.<String, List<TimeAggregation>>builder().putAll(result).build();
   }
 
-  private void getSourceToFilterExpressionMap(Filter filter,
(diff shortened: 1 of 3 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -323,22 +323,34 @@ private void buildSourceToOrderByExpressionMaps() {
...
+      Filter filter, Map<String, List<Expression>> sourceToFilterExpressionMap) {
     if (Filter.getDefaultInstance().equals(filter)) {
       return;
     }
 
     if (filter.hasLhs()) {
       // Assuming RHS never has columnar expressions.
       Expression lhs = filter.getLhs();
-      Map<String, List<Expression>> map = getDataSourceToExpressionMap(List.of(lhs));
-      sourceToExpressionMap.putAll(map);
```
skjindal93, line 336: `putAll` is bad! It does not merge the map, it replaces the existing keys in the map

tim-mwangi (approved):

PR #70, opened 2021-01-19
fix(entities): selections on eds, and filters/order by on qs

skjindal93:
## Description
Entities request was failing if the selections were both on EDS and QS, but filters and order by are on QS. The reason for failure being the AndNode/OrNode in `FilterOptimizingVisitor`assumes that the children can only be DataFetcherNode and nothing else, but since we were pushing pagination down to QS, it also had a `PaginateOnlyNode` on top of `DataFetcherNode`

### Testing
Added unit tests

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/FilterOptimizingVisitor.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/FilterOptimizingVisitor.java
@@ -110,12 +110,22 @@ public class FilterOptimizingVisitor implements Visitor<QueryNode> {
 
   @Override
   public QueryNode visit(SelectionNode selectionNode) {
-    return selectionNode;
+    QueryNode childNode = selectionNode.getChildNode().acceptVisitor(this);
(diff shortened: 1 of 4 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/FilterOptimizingVisitor.java
```diff
@@ -110,12 +110,22 @@ public QueryNode visit(OrNode orNode) {
 
   @Override
   public QueryNode visit(SelectionNode selectionNode) {
-    return selectionNode;
+    QueryNode childNode = selectionNode.getChildNode().acceptVisitor(this);
```
tim-mwangi, line 113: There are no unit tests for this class?

skjindal93: adding those just now, as I type this

avinashkolluru (dismissed):

avinashkolluru (approved):

PR #77, opened 2021-01-25
perf(entities): fetch total directly

skjindal93:
## Description
Total number of entities can be fetched directly from the data store, if pagination is pushed down to the data store

### Testing
Added unit tests. Tested end to end

### Checklist:

### DIFF
--- a/gateway-service-impl/build.gradle.kts
+++ b/gateway-service-impl/build.gradle.kts
@@ -25,8 +25,8 @@ dependencies {
 
   implementation("org.hypertrace.core.query.service:query-service-client:0.5.2")
   implementation("org.hypertrace.core.attribute.service:attribute-service-client:0.9.3")
-  implementation("org.hypertrace.entity.service:entity-service-client:0.6.0-SNAPSHOT")
-  implementation("org.hypertrace.entity.service:entity-service-api:0.6.0-SNAPSHOT")
+  implementation("org.hypertrace.entity.service:entity-service-client:0.5.6")
+  implementation("org.hypertrace.entity.service:entity-service-api:0.5.6")
   implementation("org.hypertrace.core.grpcutils:grpc-context-utils:0.3.2")
   implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.19")
 
--- a/gateway-service-impl/build.gradle.kts
+++ b/gateway-service-impl/build.gradle.kts
@@ -25,8 +25,8 @@ dependencies {
 
   implementation("org.hypertrace.core.query.service:query-service-client:0.5.2")
   implementation("org.hypertrace.core.attribute.service:attribute-service-client:0.9.3")
-  implementation("org.hypertrace.entity.service:entity-service-client:0.5.0")
-  implementation("org.hypertrace.entity.service:entity-service-api:0.5.0")
+  implementation("org.hypertrace.entity.service:entity-service-client:0.6.0-SNAPSHOT")
+  implementation("org.hypertrace.entity.service:entity-service-api:0.6.0-SNAPSHOT")
   implementation("org.hypertrace.core.grpcutils:grpc-context-utils:0.3.2")
   implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.19")
 
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
@@ -18,6 +18,9 @@ public class QueryRequestUtil {
 
   public static final String DATE_TIME_CONVERTER = "dateTimeConvert";
 
+  private static final String COUNT_FUNCTION_NAME = "COUNT";
+  private static final String DISTINCTCOUNT_FUNCTION_NAME = "DISTINCTCOUNT";
+
   public static Filter createBetweenTimesFilter(String columnName, long lower, long higher) {
     return Filter.newBuilder()
         .setOperator(Operator.AND)
@@ -102,7 +105,7 @@ public class QueryRequestUtil {
     return Expression.newBuilder()
         .setFunction(
             Function.newBuilder()
-                .setFunctionName("COUNT")
+                .setFunctionName(COUNT_FUNCTION_NAME)
                 .addArguments(createColumnExpression(columnName)))
         .build();
   }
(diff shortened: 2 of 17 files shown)


gateway-service-impl/build.gradle.kts
```diff
@@ -25,8 +25,8 @@ dependencies {
 
   implementation("org.hypertrace.core.query.service:query-service-client:0.5.2")
   implementation("org.hypertrace.core.attribute.service:attribute-service-client:0.9.3")
-  implementation("org.hypertrace.entity.service:entity-service-client:0.5.0")
-  implementation("org.hypertrace.entity.service:entity-service-api:0.5.0")
+  implementation("org.hypertrace.entity.service:entity-service-client:0.6.0-SNAPSHOT")
```
skjindal93, line 28: Depends on https://github.com/hypertrace/entity-service/pull/76

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
```diff
@@ -174,4 +179,35 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
...
+  public long getTotal(EntitiesRequestContext requestContext, EntitiesRequest entitiesRequest) {
+    EntityQueryRequest.Builder builder =
+        EntityQueryRequest.newBuilder()
+            .setEntityType(entitiesRequest.getEntityType())
+            .setFilter(
+                EntityServiceAndGatewayServiceConverter.convertToEntityServiceFilter(
+                    entitiesRequest.getFilter()));
+
+    // add time filter for supported scope
+    EntityServiceAndGatewayServiceConverter.addBetweenTimeFilter(
```
avinashkolluru, line 193: Is there a time filter for EntityDataService queries?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -121,13 +115,8 @@ protected static EntityResponse union(List<EntityResponse> entityResponses) {
...
                 .map(EntityResponse::getEntityFetcherResponse)
                 .collect(Collectors.toList()));
-    Set<EntityKey> entityKeys =
-        entityResponses.parallelStream()
-            .map(EntityResponse::getEntityKeys)
-            .reduce(Sets::union)
-            .orElse(Collections.emptySet());
 
-    return new EntityResponse(entityFetcherResponse, entityKeys);
+    return new EntityResponse(entityFetcherResponse, entityFetcherResponse.size());
```
avinashkolluru, line 119: Can there be a case where the child DataFetcherNode set the total, but we are overriding the size again here?

skjindal93: The union would happen when there is an OR node. 
The OR node would be there where there is an OR filter across different data sources. 
If there is an OR node across different data sources, the DataFetcherNode won't fetch total from the data store, since it only fetches total when pagination can be pushed down to the data store. 
The pagination will only be pushed down to the data store, when the filters are on the same data source

skjindal93: Unless, the optimization for OR and AND for merging same data fetcher nodes goes for a toss

skjindal93: I can be safe, and compute total, only if not set by the child data fetcher nodes, by using a `Long`, instead of `long`

avinashkolluru: Isn't that safer?

skjindal93: Yes. But, then if there are multiple data fetcher nodes, whose total to prefer?

skjindal93: We can add an error LOG, if there are more than one data fetcher nodes for the same source, under OR and AND, so that we know that something is weird

avinashkolluru: Ok

skjindal93: ERROR log might be tricky. Let me add unit tests for OR and AND nodes

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcherTests.java
```diff
@@ -187,6 +190,50 @@ public void test_getTimeAggregatedMetrics() {
         });
   }
 
+  @Nested
```
tim-mwangi, line 193: Why do we need this? Why not just have TotalEntitiesTest class?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
```diff
@@ -107,6 +107,15 @@ public static Expression createCountByColumnSelection(String columnName) {
         .build();
   }
 
+  public static Expression createDistinctCountByColumnSelection(String columnName) {
+    return Expression.newBuilder()
+        .setFunction(
+            Function.newBuilder()
+                .setFunctionName("DISTINCTCOUNT")
```
tim-mwangi, line 114: make static final.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcherTests.java
skjindal93: The method `getTotal` lies in the same class as `EntityDataServiceEntityFetcher`

Nested annotation makes the unit test more beautiful and structured, such that any unit tests related to total would go inside this Nested class, and won't be spread and mixed up across unit tests

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
skjindal93: added unit tests

tim-mwangi (approved):

avinashkolluru (approved):

PR #78, opened 2021-02-05
refactor: using micrometer based metrics and consistent with its names

kotharironak:
## Description
- Remove usage of deprecated methods from Platformregistry metrics library in support of using micrometer library
- Using prefix - `hypertrace` - for all metrics for consistency

### Testing
Tested locally using docker-compose (see notes below)

### Checklist:

### Documentation
will update the existing documentation referring to metrics and dashboards - https://github.com/hypertrace/hypertrace/tree/main/kubernetes/monitoring

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
@@ -1,11 +1,13 @@
 package org.hypertrace.gateway.service.entity;
 
-import com.codahale.metrics.Timer;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import io.micrometer.core.instrument.Timer;
+import java.time.Duration;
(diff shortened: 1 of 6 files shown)


kotharironak:
Verified locally with hypertrace-service locally build image using in docker-compose. Captured metrics,

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -1,11 +1,13 @@
 package org.hypertrace.gateway.service.entity;
 
-import com.codahale.metrics.Timer;
```
tim-mwangi, line 3: yeees! can you remove the dependency on this as well in build.gradle.kts

kotharironak: Done

tim-mwangi (approved):

PR #83, opened 2021-03-16
feat: add separate group and row limits

aaron-steinfeld:
## Description
Previously, the limit on an explorer request had different semantics depending on the rest of the request. It indicated the number of result rows if there were no groups requested, but if groups were requested, it applied there instead. Additionally, if a "rest" group was asked for, we could get <limit> groups + 1 (for rest).

The new behavior adds a `group_limit` that applies to the number of groups, _excluding_ the rest group. The top level `limit` (and offset) on the other hand always applies to the number of rows. If the group_limit is unset, `limit` will be used as `group_limit` as well.  The request will never return more rows than the top level limit (except temporarily, to satisfy backwards compatibility described below).

Backwards compatibility:
For temporary backwards compatibility with the previous limit semantics we add the following scenarios:
- If no `group_limit` is set on a grouped time series request, we use provided `limit` as a `group_limit` and expand the effective limit to 10000 to return the full time series. Once this compatibility is removed, such a request will only return `limit` rows and up to `limit` groups.
- If no `group_limit` is set on a grouped aggregation request _and_ a rest group is requested, we expand the effective limit by 1 to leave room for a rest group. Once this compatibility is removed, such a request will return up to `limit` groups, and a rest group only if the limit is not reached by the named groups.

### Testing
Good amount of e2e testing, including adding component level tests for new scenarios and verifying backwards compatibility with old tests.

### Checklist:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -116,10 +114,11 @@ QueryRequest buildQueryRequest(
 
   private Iterator<ResultSetChunk> executeQuery(
       ExploreRequestContext context, QueryRequest queryRequest) {
-    if (getLogger().isDebugEnabled()) {
+    // FIXME: revert to debug before merging
```
aaron-steinfeld, line 117: ah, of course I forget these. stupid logging for the rests. hang on...

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -212,16 +212,16 @@ private ExploreRequest createRequest(
...
 
-  private List<String> getExcludedValues(
+  private Set<String> getExcludedValues(
       String columnName, ExploreResponse.Builder originalResponse) {
     // GroupBy only supports columns expressions for now.
     return originalResponse.getRowBuilderList().stream()
         .map(rowBuilder -> rowBuilder.getColumnsMap().get(columnName))
         .map(Value::getString)
-        .collect(Collectors.toUnmodifiableList());
+        .collect(ImmutableSet.toImmutableSet());
```
aaron-steinfeld, line 221: this was a minor bug that came up in the tests - we were allowing dupes in the exclusion filter. Using a guava immutable set (which maintains order) dedupes but still keeps order predictable.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandlerTest.java
```diff
@@ -12,105 +11,14 @@
...
-  private final Row.Builder builder9 = createRowBuilder("blahService", 6, 3.7);
-  private final Row.Builder builder10 = createRowBuilder("hellService", 6, 0.7);
-  private final Row.Builder builder11 = createRowBuilder("bottleService", 7, 9.7);
-  private final Row.Builder builder12 = createRowBuilder("listeningService", 10, 1.7);
-  private final Row.Builder builder13 = createRowBuilder("wowService", 10, 6.7);
-  private final Row.Builder builder14 = createRowBuilder("hahaService", 2, 90.7);
-
-  @Test
-  public void testSortAndPaginateRowBuilders() {
-    List<Row.Builder> rowBuilders = createRowBuildersList();
```
aaron-steinfeld, line 43: This code under test was removed as we no longer limit within intervals.

tim-mwangi (approved):

aaron-steinfeld:
Bintray's been having issues all day - snyk-scan is the only job that doesn't use a cache.

PR #85, opened 2021-04-05
chore: added code style

skjindal93:
## Description
Added code style


avinashkolluru (approved):

jcchavezs:
I think a next step here would be to fail the CI if the code style isn't met.

skjindal93:
> I think a next step here would be to fail the CI if the code style isn't met.

The code style plugin automatically takes care of it. It runs code style checks as part of gradle build step

PR #87, opened 2021-04-15
feat: Api for log event

rish691:
Adds LogEvents api

https://github.com/hypertrace/hypertrace/issues/224


gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "org.hypertrace.gateway.service.v1.log.events";
+
+package org.hypertrace.gateway.service.v1.log.event;
+
+import "org/hypertrace/gateway/service/v1/gateway_query.proto";
+
+message LogEventRequest {
```
aaron-steinfeld, line 10: can we add the limit, order, offset etc. params

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
```diff
@@ -0,0 +1,22 @@
...
+  repeated org.hypertrace.gateway.service.v1.common.Expression selection = 2;
+}
+
+message LogEvent {
+  map<string, org.hypertrace.gateway.service.v1.common.Value> attributes = 1;
+}
+
+message LogEventResponse {
+  repeated LogEvent logEvents = 1;
+  int32 total = 2;
```
aaron-steinfeld, line 21: My understanding is this needs to be fetched separately, so we've kind of retrofitted an "includeTotal" flag on other APIs. Since this is a new one, it may be worth looking at those and if it makes sense, splitting fetching total into its own API.

rish691: Why does `total` needs to be fetched separately? and if that is added, should the existing api also return `total`?

aaron-steinfeld: It was a perf optimization. In many contexts (for example, showing the log events in the sequence diagram), we don't need the total - but under the hood, the gateway service had to make a separate query (which I believe was sequential, although that could be fixed) to fetch it and it slowed down the queries that didn't need it. By keeping it separate, we can fetch it in parallel if needed and leave it be if it's not requested. @skjindal93 may have more context.

skjindal93: > Why does `total` needs to be fetched separately? and if that is added, should the existing api also return `total`?

I think there are 2 questions here

1. Why does `total` needs to be fetched separately?

If the request has support for **limit** and **offset**, which usually gets pushed down to the data store, you cannot count number of log events in the response and call it as **total** (since the count of log events would just be bounded by the **limit**)
A separate request needs to be made to get the **total** value, by just passing the filters sent in the request (_And, that's why I personally prefer to have a separate total request API entirely_)

2. What if you don't need total in the response?
That's the perf optimization which @aaron-steinfeld was mentioning about. Since, in order to fetch **total**, you need to make a separate request, but if the consumer does not need it, we should skip fetching the **total** entirely (_by either having a boolean flag like this https://github.com/hypertrace/gateway-service/blob/main/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto#L50 or a separate total request API_)

rish691: Actually, since I didn't add support for limit and order, calculating total was trivial. But, yeah with limit and order it needs another query. 
For this change, I will remove `total` from the api.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
rish691: Currently we want to support query on spanId/traceId, do we need limit and offset for it? also `order` isn't needed at the moment I guess.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
```diff
@@ -0,0 +1,21 @@
...
+  org.hypertrace.gateway.service.v1.common.Filter filter = 1;
+  repeated org.hypertrace.gateway.service.v1.common.Expression selection = 2;
+}
+
+message LogEvent {
+  map<string, org.hypertrace.gateway.service.v1.common.Value> attributes = 1;
+}
+
+message LogEventResponse {
+  repeated LogEvent logEvents = 1;
```
skjindal93, line 20: nit: snake_case in protos, instead of camelCase

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventService.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private void initMetrics() {
+    queryExecutionTimer =
+        PlatformMetricsRegistry.registerTimer(
+            "hypertrace.log.event.query.execution", ImmutableMap.of());
+  }
+
+  public LogEventResponse getSpansByFilter(RequestContext context, LogEventRequest request) {
+    //    Instant start = Instant.now();
```
tim-mwangi, line 40: remove commented out code

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventService.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  public LogEventResponse getLogEventsByFilter(RequestContext context, LogEventRequest request) {
+    Instant start = Instant.now();
+    try {
+      Map<String, AttributeMetadata> attributeMap =
+          attributeMetadataProvider.getAttributesMetadata(context, "LOG_EVENT");
+      LogEventResponse.Builder logEventResponseBuilder = LogEventResponse.newBuilder();
+
+      Collection<LogEvent> filteredLogEvents = filterLogEvents(context, request, attributeMap);
```
aaron-steinfeld, line 65: Should be declared as list since we care about list semantics

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventService.java
```diff
@@ -0,0 +1,155 @@
...
+      LOG.debug("Log Event Service Response: {}", response);
+      return response;
+    } finally {
+      queryExecutionTimer.record(
+          Duration.between(start, Instant.now()).toMillis(), TimeUnit.MILLISECONDS);
+    }
+  }
+
+  @VisibleForTesting
+  List<LogEvent> filterLogEvents(
```
aaron-steinfeld, line 79: This method name sounds like it's just filtering a passed list of events, but it's actually the entire query execution - filtering, limiting, ordering etc.

Also, visible for testing is an antipattern - if we think this is worth directly testing, it should be in its own class - else, test through the public api.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventService.java
```diff
@@ -0,0 +1,155 @@
...
+      ResultSetChunk chunk = resultSetChunkIterator.next();
+      LOG.debug("Received chunk: {}", chunk);
+
+      if (chunk.getRowCount() < 1) {
+        break;
+      }
+
+      for (Row row : chunk.getRowList()) {
+        LogEvent.Builder logEventBuilder = LogEvent.newBuilder();
+        for (int i = 0; i < chunk.getResultSetMetadata().getColumnMetadataCount(); i++) {
```
aaron-steinfeld, line 122: result metadata is not available in every chunk, only the first

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventService.java
rish691: Updated

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
rish691: Added

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
```diff
@@ -0,0 +1,28 @@
...
+option java_multiple_files = true;
+option java_package = "org.hypertrace.gateway.service.v1.log.events";
+
+package org.hypertrace.gateway.service.v1.log.event;
+
+import "org/hypertrace/gateway/service/v1/gateway_query.proto";
+
+message LogEventsRequest {
+  // todo change this to nano
+  sfixed64 start_time_nanos = 1;
```
aaron-steinfeld, line 12: hmmm - to keep this in nanos or millis? My gut is millis so this becomes more of an implementation detail (since the unit on the attribute itself will be handled generically), but I'm on the fence.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -417,8 +417,8 @@ public static Filter convertToQueryFilter(
   }
 
   public static Filter addTimeAndSpaceFiltersAndConvertToQueryFilter(
-      long startTimeMillis,
-      long endTimeMillis,
+      long startTime,
```
aaron-steinfeld, line 420: I don't believe this util cares about units, but just to be aware, others do (anything using datetimeconvert). We'll probably want to lookup the timestamp unit when we get to that point.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
rish691: Ok, I can do that. Currently I am just converting timestamp to nanos in gql request builder.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/gateway_query.proto
```diff
@@ -94,6 +94,7 @@ message Value {
   repeated double double_array = 13;
   repeated bool boolean_array = 14;
   map<string, string> string_map = 15;
+  string timestamp_unit = 16;
```
aaron-steinfeld, line 97: Why do we need this in the value? The requester has access to the metadata anyway so should have this info (which saves us from repeating the unit in every response value - we already do this for all other units)

rish691: Yeah that can also be done. But I will then need a different converter for Timestamp in gql, here https://github.com/hypertrace/hypertrace-core-graphql/pull/56#discussion_r617718178 since the convert method only has access to the Value object. 

> - we already do this for all other units)
where do we do this?

aaron-steinfeld: My response from offline:
> So GQL will need changes to support timestamps as instants in attributes - right now it only works that way for top level APIs. The value converter needs to be changed to a biconverter I believe to take in the attribute metadata. Or we may need both the Converter and Biconverter flavors if there’s some places we call it where we don’t have that metadata at hand (probably just implement both methods and have an internal implementation, for example). The 400 error is due to the serialization side. Attributes right now use an UnknownScalar type, which needs to add support for instants - right now it treats them as plain objects.

aaron-steinfeld: As to where we do this, I actually confused units and types. The units being joined in happens in the UI today - the types being joined in we do in GQL explorer results.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventsService.java
```diff
@@ -0,0 +1,172 @@
...
+import org.hypertrace.gateway.service.v1.log.events.LogEventsRequest;
+import org.hypertrace.gateway.service.v1.log.events.LogEventsResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LogEventsService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LogEventsService.class);
+
+  private static final String LOG_EVENT_SCOPE = "LOG_EVENT";
```
skjindal93, line 36: Should we add the scope here as well? https://github.com/hypertrace/attribute-service/blob/main/attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto#L9

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
```diff
@@ -0,0 +1,27 @@
...
+  int32 offset = 6;
+  int32 limit = 7;
+}
+
+message LogEvent {
+  map<string, org.hypertrace.gateway.service.v1.common.Value> attributes = 1;
+}
+
+message LogEventResponse {
+  repeated LogEvent log_events = 1;
```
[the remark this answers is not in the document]
rish691: So, that would need a parallel query I guess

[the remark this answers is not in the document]
skjindal93: Yes. That works too

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventsService.java
rish691: It is deprecated, check https://github.com/hypertrace/attribute-service/pull/89#pullrequestreview-636577006

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/log_event.proto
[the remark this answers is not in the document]
aaron-steinfeld: 👍 that's what I was thinking - separate api entirely since we have the luxury of doing it here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventsService.java
```diff
@@ -0,0 +1,172 @@
...
+
+    int offset = request.getOffset();
+    if (offset > 0) {
+      queryBuilder.setOffset(offset);
+    }
+  }
+
+  private static long convertFromMillis(long timestamp, String toUnit) {
+    if ("ns".equals(toUnit)) {
+      return timestamp * 1_000_000;
```
aaron-steinfeld, line 168: nit
```suggestion
      return Duration.ofMillis(timestamp).toNanos();
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventsService.java
```diff
@@ -0,0 +1,172 @@
...
+    Iterator<ResultSetChunk> resultSetChunkIterator =
+        queryServiceClient.executeQuery(queryRequest, context.getHeaders(), requestTimeout);
+
+    ResultSetMetadata resultSetMetadata = null;
+    while (resultSetChunkIterator.hasNext()) {
+      ResultSetChunk chunk = resultSetChunkIterator.next();
+
+      LOG.debug("Received chunk: {}", chunk);
+
+      if (chunk.getRowCount() < 1) {
```
aaron-steinfeld, line 122: I'm guessing this is from elsewhere, but I wonder what the point of this condition is? Even if we get a chunk with 0 rows, the code handles it correctly.

aaron-steinfeld (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/logevent/LogEventsService.java
rish691: without this condition `logEventResult.add(logEventBuilder.build());` would be executed

aaron-steinfeld: Isn't that inside the for loop which would never run? `for (Row row : chunk.getRowList()) {`

aaron-steinfeld (approved):

PR #90, opened 2021-06-30
Handling Clock-Skews in Distributed Traces

suddendust:
## Description

This PR is to handle clock-skews as described [here](https://github.com/hypertrace/hypertrace/issues/139). It is a work-in-progress as the proposal is yet to be approved.

Spans are changed in gateway-service as a part of the span transformation pipeline. A span transformation pipeline registers a list of handlers that transform a supplied list of spans. One such handler is ```ClockskewAdjuster.java```. I chose this design as this takes care of any new transformations we might have to apply to spans in the query layer.

The design assumes that there can be a bunch of strategies to adjust clock-skew (new heuristics can be added in future). Right now we have two strategies - A ```NoOpClockskewAdjuster``` adjuster (doesn't do any adjustment) and a ```JaegerBasedClockskewAdjuster``` (copies what Jaegar does). What adjuster to use can be controlled from the app config. Note that this PR does not handle controlling this feature from the UI. That is, once the config has been specified, all spans will then be adjusted.

This PR is still in early stages. ```JaegerBasedClockskewAdjuster.java```'s implementation is a bit terse and needs to be improved. I raised this so that everyone is updated on how we are implementing this feature.

### Testing
ToDo

### Checklist:

### Documentation
ToDo

### DIFF
--- a/gateway-service-impl/build.gradle.kts
+++ b/gateway-service-impl/build.gradle.kts
@@ -29,6 +29,8 @@ dependencies {
   // Needed by clusters snapshots
   implementation("com.fasterxml.jackson.core:jackson-annotations:2.11.1")
   implementation("com.fasterxml.jackson.core:jackson-databind:2.11.1")
+  annotationProcessor("org.projectlombok:lombok:1.18.18")
+  compileOnly("org.projectlombok:lombok:1.18.18")
 
   testImplementation("org.junit.jupiter:junit-jupiter:5.7.0")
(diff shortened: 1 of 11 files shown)


jcchavezs:
I think in this kind of fixes it is crucial to have the test cases to understand what use cases we are addressing, otherwise it is hard to follow (yeah, time is evil).

suddendust:
> I think in this kind of fixes it is crucial to have the test cases to understand what use cases we are addressing, otherwise it is hard to follow (yeah, time is evil).

PR #92, opened 2021-07-23
fix(entities): request unique set of attributes across sources

skjindal93:
## Description
A selection attribute in entities request can be present in multiple sources. That results in non deterministic response, since the attribute is present in multiple sources, it can be returned from either of the sources, leading to non determinism in the response

The fix is to request unique set of attributes across sources, based on the pivot source. The pivot source being EDS, in case of time agnostic entities request, else the pivot source being QS by default

### Testing
Added unit tests. And, tested the UI screens manually

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
@@ -1,6 +1,7 @@
 package org.hypertrace.gateway.service.entity.query;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Sets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -213,6 +214,36 @@ public class ExecutionContext {
 
   public void removePendingSelectionSourceForOrderBy(String source) {
     pendingSelectionSourcesForOrderBy.remove(source);
+  }
+
+  public void removeSelectionAttributes(String source, Set<String> attributes) {
+    if (!sourceToSelectionExpressionMap.containsKey(source)) {
+      return;
+    }
+
+    // converting to a mutable map
+    Map<String, List<Expression>> selectionExpressionMap =
+        new HashMap<>(sourceToSelectionExpressionMap);
+
+    List<Expression> expressions = selectionExpressionMap.get(source);
+    // get all the expressions which contains the attribute to be removed
+    List<Expression> removeExpressions =
+        expressions.stream()
+            .filter(
+                expression ->
+                    !Sets.intersection(ExpressionReader.extractColumns(expression), attributes)
+                        .isEmpty())
+            .collect(Collectors.toUnmodifiableList());
+
+    expressions.removeAll(removeExpressions);
+    if (expressions.isEmpty()) {
+      selectionExpressionMap.remove(source);
+    }
+
+    // converting back to immutable map
+    sourceToSelectionExpressionMap =
+        ImmutableMap.<String, List<Expression>>builder().putAll(selectionExpressionMap).build();
+    sourceToSelectionAttributeMap = buildSourceToAttributesMap(sourceToSelectionExpressionMap);
   }
 
   public Map<String, List<Expression>> getSourceToFilterExpressionMap() {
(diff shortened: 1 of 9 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -215,6 +216,36 @@ public void removePendingSelectionSourceForOrderBy(String source) {
     pendingSelectionSourcesForOrderBy.remove(source);
   }
 
+  public void removeSelectionAttributes(String source, Set<String> attributes) {
+    if (!sourceToSelectionExpressionMap.containsKey(source)) {
+      return;
+    }
+
+    // converting to a mutable map
```
avinashkolluru, line 224: Is the input map immutable? Why even have a immutable map when there are so many mutations happening here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeUtils.java
```diff
@@ -229,4 +229,30 @@ public static boolean areFiltersOnlyOnCurrentDataSource(
...
         .orElse(Collections.emptySet());
   }
+
+  /**
+   * Removes duplicate selection attributes from other sources using {@param source} as the pivot
+   * source
+   */
+  public static void removeDuplicateSelectionAttributes(
+      ExecutionContext executionContext, String source) {
+    Map<String, Set<String>> sourceToSelectionAttributeMap =
```
avinashkolluru, line 239: nit: Move the copy after the if below. Can avoid the unnecessary copy in early return cases.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -215,6 +216,36 @@ public void removePendingSelectionSourceForOrderBy(String source) {
...
+    Map<String, List<Expression>> selectionExpressionMap =
```
avinashkolluru, line 225: nit: guava lib has some fancy Maps.filterEntries, Maps.filterValues etc methods. See if they can be helpful here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
skjindal93: Yes. I was planning to keep it immutable, so that any modifications done to the map is done via through the functions exposed in `ExecutionContext`, and the consumers don't accidentally/deliberately modify the map directly

skjindal93: Though to keep things simple, I can change the input maps to mutable

avinashkolluru (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
```diff
@@ -215,6 +217,34 @@ public void removePendingSelectionSourceForOrderBy(String source) {
...
   }
 
+  public void removeSelectionAttributes(String source, Set<String> attributes) {
+    if (!sourceToSelectionExpressionMap.containsKey(source)) {
+      return;
+    }
+
+    Predicate<Expression> retainExpressionPredicate =
+        expression ->
+            Sets.intersection(ExpressionReader.extractColumns(expression), attributes).isEmpty();
```
rish691, line 227: nit: collecting all of `ExpressionReader.extractColumns(expression)` in a set and then taking the intersection with attributes would be more efficient I guess, as it would need a single pass over the attributes array

rish691 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionContext.java
skjindal93: That would be more trickier to maintain the fact, which expression lead to a set of corresponding attributes. Because, we need those expressions which lie outside of the intersection.

Basically a set difference, but with a custom comparator. I can take a look at it separately, if there is a cleaner way to do that

PR #93, opened 2021-07-30
fix: conversion of time interval to time column expression

GurtejSohi:
## Description
Currently, while converting the time interval to time column expression, we convert the interval into `<timeSize>:SECONDS`. This is passed to query service(Pinot) and it expects the time size portion to be an int. So, if it's outside the integer limit, it throws an error. Integer.MAX_VALUE(2147483647) seconds corresponds to ~24.85 days. So, if you specify the interval for 25 days or more, it will throw an error.

Fix: 
If the interval is outside the integer limit, increase the time unit (from Seconds to Hours) and accordingly, decrease the time size portion so that the time size fits in the integer limit.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
@@ -1,6 +1,7 @@
 package org.hypertrace.gateway.service.common.converters;
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import org.hypertrace.core.query.service.api.ColumnIdentifier;
 import org.hypertrace.core.query.service.api.Expression;
 import org.hypertrace.core.query.service.api.Filter;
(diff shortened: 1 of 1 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryRequestUtil.java
```diff
@@ -116,14 +117,22 @@ public static Expression createDistinctCountByColumnSelection(String columnName)
   }
 
   public static Expression createTimeColumnGroupByExpression(String timeColumn, long periodSecs) {
+    // pinot expects the time size portion to be an int. So, if it's outside the integer limit, we
```
aaron-steinfeld, line 120: All of this logic really belongs in query service, since gateway shouldn't even know Pinot is the storage provider. That's obviously a bigger change though, so for now I'm curious - why hours and not just minutes? Seems like that's more than sufficient and preserves a lot more granularity

GurtejSohi: Actually, I realized it later that this problem happens when we use Milliseconds as the time unit. But here, we're using Seconds as time unit - so we should be fine here without this change. Sorry for the confusion!

PR #94, opened 2021-08-13
chore: limit direct memory

skjindal93:


varsha-abhinandan (approved):

aaron-steinfeld (commented):
out of curiosity do we know the _why_ here?  Who's using direct memory, since it's not our code - Is it grpc/netty?

helm/values.yaml
```diff
@@ -24,7 +24,7 @@ service:
 
 nodeLabels: {}
 
-javaOpts: "-XX:InitialRAMPercentage=50.0 -XX:MaxRAMPercentage=75.0"
+javaOpts: "-XX:InitialRAMPercentage=50.0 -XX:MaxRAMPercentage=75.0 -XX:MaxDirectMemorySize=128M"
```
varsha-abhinandan, line 27: Wondering if there's an easy way to figure out the services which don't have this config set ? Might help avoid some future OOMs

skjindal93:
> out of curiosity do we know the _why_ here? Who's using direct memory, since it's not our code - Is it grpc/netty?

Didn't investigate that (_rather didn't know how to investigate that_)

skjindal93:
Though we need this, to avoid the service running OOM frequently

aaron-steinfeld:
> Though we need this, to avoid the service hitting OOM frequently

Increasing memory avoids running out of memory, sure :) but it's more (or just as) important to understand why we're running out of memory - is there anyone investigating that?

skjindal93:
> > Though we need this, to avoid the service hitting OOM frequently
> 
> Increasing memory avoids running out of memory, sure :) but it's more (or just as) important to understand why we're running out of memory - is there anyone investigating that?

Increasing memory? In fact, we are limiting the direct memory, otherwise, it was eating up the entire available memory

aaron-steinfeld:
> > > Though we need this, to avoid the service hitting OOM frequently
> > 
> > 
> > Increasing memory avoids running out of memory, sure :) but it's more (or just as) important to understand why we're running out of memory - is there anyone investigating that?
> 
> Increasing memory? In fact, we are limiting the direct memory, otherwise, it was eating up the entire available memory



> > > Though we need this, to avoid the service hitting OOM frequently
> > 
> > 
> > Increasing memory avoids running out of memory, sure :) but it's more (or just as) important to understand why we're running out of memory - is there anyone investigating that?
> 
> Increasing memory? In fact, we are limiting the direct memory, otherwise, it was eating up the entire available memory

er, right, sorry - but my point is if we don't understand how we're using it, we don't really understand what the implications of limiting it are - are we spiking cpu due to the memory constraints, are we affecting maximum connection or TTL values, etc.

rish691:
> Didn't investigate that (rather didn't know how to investigate that)

We could try using tools like visual vm or swiss java knife to investigate this (or maybe some free apm tool!)

helm/values.yaml
rish691: we mostly override `javaOpts` for all services in a private repo, we could just add this over there.

aaron-steinfeld (approved):
approving,  but I think there's more to be done to investigate how we can tune grpc and its memory usage.

PR #96, opened 2021-08-17
fix(converter): add missing long -> timestamp converter

skjindal93:


gateway-service-api/build.gradle.kts
```diff
@@ -47,13 +47,13 @@ sourceSets {
...
-  api("com.google.api.grpc:proto-google-common-protos:1.18.1")
-  api("io.grpc:grpc-stub:1.36.1")
+  api("io.grpc:grpc-protobuf:1.39.0")
+  api("com.google.api.grpc:proto-google-common-protos:2.3.2")
+  api("io.grpc:grpc-stub:1.39.0")
   api("javax.annotation:javax.annotation-api:1.3.2")
 
   constraints {
-    api("com.google.guava:guava:30.0-jre") {
+    api("com.google.guava:guava:30.1.1-jre") {
```
aaron-steinfeld, line 56: this constraint shouldn't even exist as far as I can tell - we shouldn't be specifying a guava version in our api unless we're using it.

gateway-service-impl/build.gradle.kts
```diff
@@ -11,28 +11,27 @@ tasks.test {
...
-  implementation("com.fasterxml.jackson.core:jackson-databind:2.11.1")
+  implementation("com.fasterxml.jackson.core:jackson-annotations:2.12.4")
+  implementation("com.fasterxml.jackson.core:jackson-databind:2.12.4")
 
   testImplementation("org.junit.jupiter:junit-jupiter:5.7.0")
   testImplementation("org.mockito:mockito-core:3.9.0")
   testImplementation("org.mockito:mockito-inline:3.9.0")
   testImplementation("org.apache.logging.log4j:log4j-slf4j-impl:2.14.0")
-  testImplementation("io.grpc:grpc-netty:1.37.0")
+  testImplementation("io.grpc:grpc-netty:1.39.0")
```
aaron-steinfeld, line 36: tests should not be using netty 🙄  but probably out of scope

aaron-steinfeld (approved):

gateway-service-impl/build.gradle.kts
skjindal93: I actually tried removing it, but there was a unit test which failed, so I put it back

gateway-service-api/build.gradle.kts
skjindal93: I tried removing it, and changed `ImmutableSet.toImmutableSet()` to `Collectors.toUnmodifiableSet()` at some places.  And, the tests failed again

aaron-steinfeld: that indicates issues - if we're using it, then we should be declaring it, not constraining it. And our api shouldn't have any code, anyway - looks like the _impl_ is the one using it and not declaring it - but instead depending on the api's constraint.

gateway-service-impl/build.gradle.kts
aaron-steinfeld: Again, indicates issues - that means we're actually spinning up transport in a unit test, which we shouldn't be doing.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/LongToAttributeKindConverter.java
```diff
@@ -26,6 +26,11 @@ public Value doConvert(Long value, AttributeKind attributeKind, Value.Builder va
         valueBuilder.setString(value.toString());
         return valueBuilder.build();
 
+      case TYPE_TIMESTAMP:
```
jcchavezs, line 29: Tip: This is a PR that mostly includes version updates and unrelated changes, I would suggest in this kind of PRs we point out what is the important change. I guess this is the important change in this PR.

PR #101, opened 2021-09-13
fix: handle the avg rate function in series request

kotharironak:
Ref issue : https://github.com/hypertrace/hypertrace/issues/285
@sarthak77  you can close your continue on this.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
@@ -20,7 +20,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.commons.lang3.StringUtils;
-import org.hypertrace.core.attribute.service.v1.AttributeKind;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.query.service.api.ColumnMetadata;
 import org.hypertrace.core.query.service.api.Expression;
@@ -422,9 +421,9 @@ public class QueryServiceEntityFetcher implements IEntityFetcher {
               QueryAndGatewayDtoConverter.convertQueryValueToGatewayValue(
                   row.getColumn(idColumns.size()));
           if (value.getValueType() == ValueType.STRING) {
-            long time = Long.parseLong(value.getString());
-            long endTime = time + TimeUnit.SECONDS.toMillis(periodSecs);
-            intervalBuilder.setStartTimeMillis(time);
+            long startTime = Long.parseLong(value.getString());
+            long endTime = startTime + TimeUnit.SECONDS.toMillis(periodSecs);
+            intervalBuilder.setStartTimeMillis(startTime);
             intervalBuilder.setEndTimeMillis(endTime);
 
             for (int i = idColumns.size() + 1;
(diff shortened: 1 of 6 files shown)


gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
```diff
@@ -81,6 +91,167 @@ public void setup() {
...
+            .setOffset(0)
+            .setLimit(10000)
+            .build();
+
+    List<ResultSetChunk> resultSetChunks =
+        List.of(
+            getResultSetChunk(
+                List.of("API.apiId", "timeColumn", "SUM_API.numCalls", "AVGRATE_API.numCalls"),
+                new String[][] {
+                  {"apiId1", "10000", "34", "43"},
```
kotharironak, line 175: The last two columns will be the same till here  (unless the time range is different) as `AVGRATE_API` internally asking SUM from Pinot handler.
```
{"apiId1", "10000", "34", "34"},
{"apiId2", "20000", "34", "34"}
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -421,14 +421,16 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
               QueryAndGatewayDtoConverter.convertQueryValueToGatewayValue(
                   row.getColumn(idColumns.size()));
           if (value.getValueType() == ValueType.STRING) {
-            long time = Long.parseLong(value.getString());
-            intervalBuilder.setStartTimeMillis(time);
-            intervalBuilder.setEndTimeMillis(time + TimeUnit.SECONDS.toMillis(periodSecs));
+            long startTime = Long.parseLong(value.getString());
```
aaron-steinfeld, line 424: A bit of an aside: The fact that avg rate was implemented in gateway has always been tech debt - that responsibility to evaluate a function belongs to query service (which becomes even more relevant as we consider adding a metric store). Any thought on level of effort to moving it? It may make sense to do as part of a broader effort to remove _any_ pinot knowledge from gateway - of which there's quite a bit today.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
```diff
@@ -421,14 +421,16 @@ public EntityFetcherResponse getTimeAggregatedMetrics(
...
+            long endTime = startTime + TimeUnit.SECONDS.toMillis(periodSecs);
+            intervalBuilder.setStartTimeMillis(startTime);
+            intervalBuilder.setEndTimeMillis(endTime);
 
             for (int i = idColumns.size() + 1;
                 i < chunk.getResultSetMetadata().getColumnMetadataCount();
                 i++) {
               ColumnMetadata metadata = chunk.getResultSetMetadata().getColumnMetadata(i);
+              // this where we have to add the values.
```
aaron-steinfeld, line 433: What does this comment mean? Can we either remove or clarify?

sarthak77: CC @kotharironak

sarthak77: @aaron-steinfeld mostly this section of code is for adding the value to the response (`resultmap` in above code) which will be returned. Until here we were extracting some meta information about the data but at this point we find out the value from time aggregation (`row.getColumn(i)` in above code).

kotharironak: @sarthak77 This comment is not required. I was just highlighting in draft PR where you have to make changes.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java
kotharironak: This makes sense. I guess it shouldn't be a big effort for avg_rate. We have to clean up the translation mapping from avg_rate to SUM and move the calculation to query service. I have created the issue for this specific case - https://github.com/hypertrace/hypertrace/issues/296, and we will take it up as follow up. I need to scan for other pinot deps and will move to its own ticket.

kotharironak (commented):
lgtm

rish691 (approved):

aaron-steinfeld (approved):

Issue #105, opened 2021-10-18
Cleaning up average rate code to be directly used from query service. 


sarthak77:
Ref issue : https://github.com/hypertrace/hypertrace/issues/296

PR #106, opened 2021-10-19
cleaning up average rate implementation from gateway service

sarthak77:
Ref issue : https://github.com/hypertrace/gateway-service/issues/105

As a part of this PR, the following changes are being made : 

1. Since AVG_RATE support added in QS, removing the transformations from GS
2. Adding support for literal / rate interval to be in ISO string format
3. Maintaining support for literal / rate interval to be in long for backward compatibility

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
@@ -19,7 +19,6 @@ import org.hypertrace.gateway.service.common.AttributeMetadataProvider;
 import org.hypertrace.gateway.service.common.util.AttributeMetadataUtil;
 import org.hypertrace.gateway.service.entity.EntitiesRequestContext;
 import org.hypertrace.gateway.service.entity.config.TimestampConfigs;
-import org.hypertrace.gateway.service.v1.common.FunctionType;
 import org.hypertrace.gateway.service.v1.entity.EntitiesRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -165,38 +164,48 @@ public class EntityServiceAndGatewayServiceConverter {
             .setFunctionName(function.getFunction().name())
             .setAlias(function.getAlias());
 
-    if (function.getFunction() == FunctionType.PERCENTILE) {
-      org.hypertrace.gateway.service.v1.common.Expression percentileExp =
-          function.getArgumentsList().stream()
-              .filter(org.hypertrace.gateway.service.v1.common.Expression::hasLiteral)
-              .findFirst()
-              .orElseThrow(); // Should have validated arguments using PercentileValidator
+    switch (function.getFunction()) {
+      case AVGRATE:
+        {
+          throw new IllegalArgumentException("Doesn't support AVGRATE on entity queries");
+        }
+      case PERCENTILE:
+        {
+          org.hypertrace.gateway.service.v1.common.Expression percentileExp =
+              function.getArgumentsList().stream()
+                  .filter(org.hypertrace.gateway.service.v1.common.Expression::hasLiteral)
+                  .findFirst()
+                  .orElseThrow(); // Should have validated arguments using PercentileValidator
 
-      long percentile = percentileExp.getLiteral().getValue().getLong();
-      String functionName = function.getFunction().name() + percentile;
-      builder.setFunctionName(functionName).setAlias(function.getAlias());
+          long percentile = percentileExp.getLiteral().getValue().getLong();
+          String functionName = function.getFunction().name() + percentile;
+          builder.setFunctionName(functionName).setAlias(function.getAlias());
 
-      // Adds only the argument that is a literal identifier. This will bring the required nth
-      // percentile.
-      List<org.hypertrace.gateway.service.v1.common.Expression> columns =
-          function.getArgumentsList().stream()
-              .filter(org.hypertrace.gateway.service.v1.common.Expression::hasColumnIdentifier)
-              .collect(Collectors.toList());
-      columns.forEach(e -> builder.addArguments(convertToEntityServiceExpression(e)));
-    } else {
-      builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
+          // Adds only the argument that is a literal identifier. This will bring the required nth
+          // percentile.
+          List<org.hypertrace.gateway.service.v1.common.Expression> columns =
+              function.getArgumentsList().stream()
+                  .filter(org.hypertrace.gateway.service.v1.common.Expression::hasColumnIdentifier)
+                  .collect(Collectors.toList());
+          columns.forEach(e -> builder.addArguments(convertToEntityServiceExpression(e)));
+          break;
+        }
+      default:
+        {
+          builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
 
-      if (function.getArgumentsCount() > 0) {
-        function
-            .getArgumentsList()
-            .forEach(
-                e -> {
-                  // Health expressions are treated differently so ignore them.
-                  if (!e.hasHealth()) {
-                    builder.addArguments(convertToEntityServiceExpression(e));
-                  }
-                });
-      }
+          if (function.getArgumentsCount() > 0) {
+            function
+                .getArgumentsList()
+                .forEach(
+                    e -> {
+                      // Health expressions are treated differently so ignore them.
+                      if (!e.hasHealth()) {
+                        builder.addArguments(convertToEntityServiceExpression(e));
+                      }
+                    });
+          }
+        }
     }
     return builder.build();
   }
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
@@ -1,8 +1,7 @@
 package org.hypertrace.gateway.service.common.converters;
 
-import static org.hypertrace.gateway.service.common.util.QueryExpressionUtil.convertLiteralExpressionToIsoDurationString;
-
 import com.google.common.base.Strings;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -331,6 +330,26 @@ public class QueryAndGatewayDtoConverter {
     return LiteralConstant.newBuilder()
         .setValue(convertGatewayValueToQueryApiValue(literal.getValue()))
         .build();
+  }
+
+  private static org.hypertrace.gateway.service.v1.common.Expression
+      convertLiteralExpressionToIsoDurationString(
+          org.hypertrace.gateway.service.v1.common.Expression expression) {
+    // expression can be either long or an iso string
+    if (expression.getLiteral().getValue().getValueType()
+        == org.hypertrace.gateway.service.v1.common.ValueType.LONG) {
+      return org.hypertrace.gateway.service.v1.common.Expression.newBuilder()
+          .setLiteral(
+              org.hypertrace.gateway.service.v1.common.LiteralConstant.newBuilder()
+                  .setValue(
+                      org.hypertrace.gateway.service.v1.common.Value.newBuilder()
+                          .setString(
+                              Duration.ofSeconds(expression.getLiteral().getValue().getLong())
+                                  .toString())
+                          .setValueType(org.hypertrace.gateway.service.v1.common.ValueType.STRING)))
+          .build();
+    }
+    return expression;
   }
 
   private static Function convertToQueryFunction(
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/QueryExpressionUtil.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/QueryExpressionUtil.java
@@ -1,6 +1,5 @@
 package org.hypertrace.gateway.service.common.util;
 
-import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -148,22 +147,5 @@ public class QueryExpressionUtil {
     return delta == 0
         ? timeMillis
         : alignToNext ? timeMillis + (periodMillis - delta) : timeMillis - delta;
-  }
-
-  public static Expression convertLiteralExpressionToIsoDurationString(Expression expression) {
-    // expression can be either long or an iso string
-    if (expression.getLiteral().getValue().getValueType() == ValueType.LONG) {
-      return Expression.newBuilder()
-          .setLiteral(
-              LiteralConstant.newBuilder()
-                  .setValue(
-                      Value.newBuilder()
-                          .setString(
-                              Duration.ofSeconds(expression.getLiteral().getValue().getLong())
-                                  .toString())
-                          .setValueType(ValueType.STRING)))
-          .build();
-    }
-    return expression;
   }
 }
--- a/gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/validators/function/AvgRateValidatorTest.java
+++ b/gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/validators/function/AvgRateValidatorTest.java
@@ -384,28 +384,6 @@ public class AvgRateValidatorTest {
     }
 
     @Test
-    public void avgRateFunctionWithUnsetPeriod_shouldThrowIllegalArgException() {
-      FunctionExpression functionExpression =
-          FunctionExpression.newBuilder()
-              .setFunction(FunctionType.AVGRATE)
-              .setAlias("avg-rate-alias")
-              .addArguments(
-                  Expression.newBuilder()
-                      .setColumnIdentifier(
-                          ColumnIdentifier.newBuilder().setColumnName("test-column")))
-              .addArguments(Expression.newBuilder().setLiteral(LiteralConstant.newBuilder()))
-              .build();
-
-      var avgRateFunctionValidator = new AvgRateValidator();
-
-      Assertions.assertThrows(
-          IllegalArgumentException.class,
-          () -> {
-            avgRateFunctionValidator.validate(functionExpression);
-          });
-    }
-
-    @Test
     public void avgRateFunctionWithZeroPeriod_shouldThrowIllegalArgException() {
       FunctionExpression functionExpression =
           FunctionExpression.newBuilder()


gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
```diff
@@ -144,7 +144,8 @@ public void testBaselineEntitiesForAggregatesForAvgRateFunction() {
     BaselineEntity baselineEntity = baselineResponse.getBaselineEntityList().get(0);
     // verify the baseline for AVG RATE (medianValue/60)
     Assertions.assertEquals(
-        1.0, baselineEntity.getBaselineAggregateMetricMap().get("numCalls").getValue().getDouble());
+        60.0,
```
sarthak77, line 147: Updated value directly from actual result displayed by IDE

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
```diff
@@ -192,7 +195,7 @@ public void testGetTimeAggregatedMetrics() {
                 Interval.newBuilder()
                     .setStartTimeMillis(10000)
                     .setEndTimeMillis(40000)
-                    .setValue(Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(68.0)))
+                    .setValue(Value.newBuilder().setValueType(ValueType.DOUBLE).setDouble(34.0)))
```
sarthak77, line 198: Updated value directly from actual result displayed by IDE

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -165,21 +164,7 @@ private static Function convertToQueryFunction(
             .setFunctionName(function.getFunction().name())
             .setAlias(function.getAlias());
 
-    // AVGRATE is adding a specific implementation because Pinot does not directly support this
-    // function
     switch (function.getFunction()) {
-      case AVGRATE:
```
kotharironak, line 171: At gateway service layer, the AVGRAGE argument accept literal for now.
```
function {
    function: AVGRATE
    arguments {
      columnIdentifier {
        columnName: "BACKEND.errorCount"
      }
    }
    arguments {
      literal {
        value {
          valueType: LONG
          long: 1
        }
      }
    }
```
At QS will now take it as isoDurationString as an argument. So, the argument conversion will require. Or, we will have to change the client at GraphQL layer.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -339,21 +338,7 @@ private static Function convertToQueryFunction(
             .setFunctionName(function.getFunction().name())
             .setAlias(function.getAlias());
 
-    // AVGRATE is adding a specific implementation because Pinot does not directly support this
-    // function
     switch (function.getFunction()) {
-      case AVGRATE:
-        {
```
kotharironak, line 346: Same as above - https://github.com/hypertrace/gateway-service/pull/106/files#r731584076

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/validators/function/AvgRateValidator.java
```diff
@@ -30,13 +33,13 @@ protected void validateArguments(FunctionExpression functionExpression) {
...
         case LITERAL:
-          // Need the Period to be set
+          // Need the Period to be set in ISO format
           checkArgument(
               argument.getLiteral().hasValue()
-                  && argument.getLiteral().getValue().getValueType() == ValueType.LONG,
-              "Period not set as a Long Type");
-          Long period = argument.getLiteral().getValue().getLong();
-          checkArgument(period > 0L, "Period should be > 0");
+                  && argument.getLiteral().getValue().getValueType() == ValueType.STRING,
```
kotharironak, line 39: For now, it will remain as is.

gateway-service-impl/src/test/resources/expected-test-responses/explore/simple-aggregations.json
```diff
@@ -16,7 +16,7 @@
         },
         "RATE#results/rateLatency:Api.Trace|duration": {
           "valueType": "DOUBLE",
-          "double": 4774.983888888889
```
kotharironak, line 19: why is this change?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -205,6 +197,34 @@ private static Function convertToQueryFunction(
...
+            if (columns.size() > 0) {
+              columns.forEach(
+                  e ->
+                      builder.addArguments(
+                          Expression.newBuilder()
+                              .setLiteral(
+                                  LiteralConstant.newBuilder()
+                                      .setValue(
+                                          Value.newBuilder()
+                                              .setString(longToIso(e))
```
kotharironak, line 220: we can inline function `longToIso` with `Duration.ofSeconds(expression.getLiteral().getValue().getLong()).toString()`

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
kotharironak: But, why is it change?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/RequestHandlerTest.java
```diff
@@ -54,8 +54,8 @@ public void orderByExpressionsWithFunction_shouldMatchCorrespondingSelections()
                                         LiteralConstant.newBuilder()
                                             .setValue(
                                                 Value.newBuilder()
-                                                    .setValueType(ValueType.LONG)
-                                                    .setLong(30))))))
+                                                    .setValueType(ValueType.STRING)
```
kotharironak, line 57: I think none of the existing tests will change. But, we will have to add new tests where requests can take iso string as well.

gateway-service-impl/src/test/resources/expected-test-responses/explore/simple-aggregations.json
kotharironak: Got it. But, wasn't it comparing the final value?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
```diff
@@ -122,22 +122,14 @@ public static Value getValueFromFunction(
...
-    } else {
-      convertedValue =
-          QueryAndGatewayDtoConverter.convertToGatewayValueForMetricValue(
-              MetricAggregationFunctionUtil.getValueTypeFromFunction(
-                  functionExpression, attributeMetadataMap),
-              attributeMetadataMap,
-              metadata,
-              column);
-    }
+    convertedValue =
```
aaron-steinfeld, line 126: nit: return directly

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
aaron-steinfeld: This seems like a red flag and caught bug. At the service level, the same arguments should give the same responses they always have for backwards compatibility.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
```diff
@@ -229,7 +230,8 @@ private FunctionExpression getFunctionExpressionForAvgRate(
             Expression.newBuilder()
                 .setLiteral(
                     LiteralConstant.newBuilder()
-                        .setValue(Value.newBuilder().setLong(60).setValueType(ValueType.LONG))))
+                        .setValue(
+                            Value.newBuilder().setString("PT1M").setValueType(ValueType.STRING))))
```
aaron-steinfeld, line 234: We should make sure to test this both ways - with an ISO string and a plain long.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
aaron-steinfeld: Test results shouldn't have changed if the implementation was correct

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/validators/function/AvgRateValidatorTest.java
```diff
@@ -316,7 +335,9 @@ public void avgRateFunctionWithEmptyAlias_shouldThrowIllegalArgException() {
                     .setLiteral(
                         LiteralConstant.newBuilder()
                             .setValue(
-                                Value.newBuilder().setLong(20L).setValueType(ValueType.LONG))))
+                                Value.newBuilder()
```
aaron-steinfeld, line 338: Make sure to keep tests support long version - for the most part, this change should only add tests because it needs to maintain existing support (may be some exceptions for this for internal things that have changed).

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/RequestHandlerTest.java
aaron-steinfeld: should have read the other comments first - I repeated this in a few other spots 😉

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -158,28 +159,19 @@ public static LiteralConstant convertToQueryLiteral(
         .build();
   }
 
+  private static String longToIso(org.hypertrace.gateway.service.v1.common.Expression expression) {
```
aaron-steinfeld, line 162: naming nit: literalSecondsToIsoString . A long isn't implicitly in seconds - let's be explicit, and also make sure that matches existing assumptions.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -205,6 +197,34 @@ private static Function convertToQueryFunction(
         {
           builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
 
+          // Backward compatibility to handle long values
+          if (function.getFunction() == FunctionType.AVGRATE) {
```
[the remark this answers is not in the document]
aaron-steinfeld: Actually, now that I look at the next file - isn't this the converter for Entity Service? We didn't add this support to ES that I saw, so this file can remain unmodified (comments here generally apply in the other converter though). That said, not sure that we ever use or would want to use avg_rate against ES - it has no concept of time, so maybe it would make sense to remove this entirely (not a bad idea considering we've changed the validator and it no longer aligns with the assumptions here).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -205,6 +197,34 @@ private static Function convertToQueryFunction(
...
+            List<org.hypertrace.gateway.service.v1.common.Expression> columns =
+                function.getArgumentsList().stream()
+                    .filter(
+                        e ->
+                            e.hasLiteral()
+                                && e.getLiteral().getValue().getValueType()
+                                    == org.hypertrace.gateway.service.v1.common.ValueType.LONG)
```
aaron-steinfeld, line 208: So as I'm currently reading this, we're special casing long args for AVG_RATE and converting them to ISO string, but if it's not a avg rate + long, it hits the default case?

The logic here is really confusing, you never want to have nested break statements inside conditions - very error prone.

What I would recommend is to maintain the avg rate case, and then pass the function expression into a normalizer that handles taking either the long or iso version and standardizing them into just the iso version. Same util can be used at validation time. Then, it can come through the standard conversion logic.

aaron-steinfeld:
@sarthak77 why are you deleting pull request comments? We should **never** be deleting comments from others.

sarthak77:
> @sarthak77 why are you deleting pull request comments? We should **never** be deleting comments from others.

@aaron-steinfeld deleting the resolved ones only 😅

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
aaron-steinfeld: That sounds like the mock should be updated, to reflect the new QS behavior, rather than the expected result.

aaron-steinfeld:
> > @sarthak77 why are you deleting pull request comments? We should **never** be deleting comments from others.
> 
> @aaron-steinfeld deleting the resolved ones only 😅

Please don't do that - it's important to leave the history. Further, it's bad etiquette to even resolve them - just respond when they're fixed. The decision to resolve is meant for the commenter, after they review the change.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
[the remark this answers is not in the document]
sarthak77: > Actually, now that I look at the next file - isn't this the converter for Entity Service? We didn't add this support to ES that I saw, so this file can remain unmodified (comments here generally apply in the other converter though). That said, not sure that we ever use or would want to use avg_rate against ES - it has no concept of time, so maybe it would make sense to remove this entirely (not a bad idea considering we've changed the validator and it no longer aligns with the assumptions here).

@kotharironak Aaron's comment which got deleted due to some ui issue on git.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
sarthak77: resolved

gateway-service-impl/src/test/resources/expected-test-responses/explore/simple-aggregations.json
sarthak77: resolved

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
sarthak77: Mock section is being used in other places too. So update over here is fine ig ?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -165,19 +166,18 @@ private static Function convertToQueryFunction(
...
-          builder.setFunctionName(FunctionType.SUM.name()).setAlias(function.getAlias());
-
-          // Adds only the argument that is a column identifier for now.
-          List<org.hypertrace.gateway.service.v1.common.Expression> columns =
-              function.getArgumentsList().stream()
-                  .filter(org.hypertrace.gateway.service.v1.common.Expression::hasColumnIdentifier)
-                  .collect(Collectors.toList());
-          columns.forEach(e -> builder.addArguments(convertToEntityServiceExpression(e)));
+          builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
+          if (function.getArgumentsCount() > 0) {
```
kotharironak, line 173: Just curious: Shouldn't there be a single argument if the validator has validated? If, that is the case, do we have to go over it all?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/validators/function/AvgRateValidatorTest.java
```diff
@@ -10,320 +10,664 @@
...
-                Expression.newBuilder()
-                    .setLiteral(
-                        LiteralConstant.newBuilder()
-                            .setValue(
-                                Value.newBuilder().setLong(20L).setValueType(ValueType.LONG))))
-            .build();
-    var avgRateFunctionValidator = new AvgRateValidator();
-    avgRateFunctionValidator.validate(functionExpression);
-  }
+  @Nested
```
kotharironak, line 18: Any reason to do nested?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/ExploreRequestValidatorTest.java
```diff
@@ -106,8 +106,8 @@ public void aggregatedSelections_shouldBeValid() {
                                         LiteralConstant.newBuilder()
                                             .setValue(
                                                 Value.newBuilder()
-                                                    .setValueType(ValueType.LONG)
-                                                    .setLong(30))))))
+                                                    .setValueType(ValueType.STRING)
```
kotharironak, line 109: This test should work as is, right?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/RequestHandlerTest.java
kotharironak: Now, you have added a test for iso duration request explicitly, can we revert this back to the existing argument (Long one)?

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/RequestHandlerTest.java
```diff
@@ -131,8 +131,8 @@ public void noChangeIfOrderByExpressionHasSameAliasAsSelection() {
...
```
kotharironak, line 134: Should be the same as prior. It should work with Long argument too.

gateway-service-impl/src/test/resources/test-requests/explore/simple-aggregations.json
```diff
@@ -64,8 +64,8 @@
           {
             "literal": {
               "value": {
-                "valueType": "LONG",
-                "long": "2"
+                "valueType": "STRING",
```
kotharironak, line 67: Should be the same as prior, right? If requires, we need to add a new request test.

kotharironak: Yes, now, you can revert back.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
kotharironak: Would suggest having mock data specific for the avg_rate.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
kotharironak: Same suggestion as @aaron-steinfeld, instead of changing assertion, can have new mock data specific to avg_rate. For this case,
```
public List<ResultSetChunk> getResultSetForAvgRate(String alias) {
    long time = Instant.parse("2020-11-14T18:40:51.902Z").toEpochMilli();

    return List.of(
        getResultSetChunk(
            List.of("SERVICE.id", "dateTimeConvert", alias),
            new String[][] {
                {"entity-1", String.valueOf(time), "0.3333333333333333"},
                {"entity-1", String.valueOf(time - 60000), "0.6666666666666666"},
                {"entity-1", String.valueOf(time - 120000), "1.0"},
                {"entity-1", String.valueOf(time - 180000), "1.3333333333333333"},
                {"entity-1", String.valueOf(time - 180000), "1.6666666666666667"},
            }));
  }
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -339,19 +340,16 @@ private static Function convertToQueryFunction(
...
-              function.getArgumentsList().stream()
-                  .filter(org.hypertrace.gateway.service.v1.common.Expression::hasColumnIdentifier)
-                  .collect(Collectors.toList());
-          columns.forEach(e -> builder.addArguments(convertToQueryExpression(e)));
+          builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
+          if (function.getArgumentsCount() > 0) {
+            function
+                .getArgumentsList()
+                .forEach(
+                    e -> builder.addArguments(convertToQueryExpression(convertLongToIsoFormat(e))));
```
kotharironak, line 351: This is functionally correct, but as only literal argument we would like to convert to iso duration string, for nit (readable), can we do as below?
```
function.getArgumentsList().stream()
              .map(e -> e.hasLiteral() ? convertLongToIsoFormat(e) : e)
              .forEach(e -> builder.addArguments(convertToQueryExpression(e)));
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
kotharironak: For readability, as only literal would like to convert to iso duration, can we do something like below?
```
function.getArgumentsList().stream()
              .map(e -> e.hasLiteral() ? convertLongToIsoFormat(e) : e)
              .forEach(e -> builder.addArguments(convertToEntityServiceExpression(e)));
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/EntityServiceAndGatewayServiceConverter.java
```diff
@@ -165,19 +166,18 @@ private static Function convertToQueryFunction(
...
         {
-          builder.setFunctionName(FunctionType.SUM.name()).setAlias(function.getAlias());
-
-          // Adds only the argument that is a column identifier for now.
-          List<org.hypertrace.gateway.service.v1.common.Expression> columns =
-              function.getArgumentsList().stream()
-                  .filter(org.hypertrace.gateway.service.v1.common.Expression::hasColumnIdentifier)
-                  .collect(Collectors.toList());
-          columns.forEach(e -> builder.addArguments(convertToEntityServiceExpression(e)));
+          builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
```
aaron-steinfeld, line 172: My suggestion would be to remove this case entirely (Can just throw, if it doesn't already by default). This logic won't work, because ES doesn't support AVGRATE - and I'm pretty sure it's never used.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/QueryExpressionUtil.java
```diff
@@ -148,4 +149,21 @@ public static long alignToPeriodBoundary(long timeMillis, long periodSecs, boole
         ? timeMillis
         : alignToNext ? timeMillis + (periodMillis - delta) : timeMillis - delta;
   }
+
+  public static Expression convertLongToIsoFormat(Expression expression) {
```
aaron-steinfeld, line 153:
```suggestion
  public static Expression tryNormalizeLiteralToIsoDurationString(Expression expression) {
```

Or something like that. There's a few different things here.

- We're still assuming it's seconds. I think that's fine, especially since it's deprecated, but if we leave "long" in the name it would be better represented by something like "literalSeconds"
- We want to accept either an isoDuration literal or a seconds literal, so I'm dropping the earlier point and just going with "normalizeLiteral"
- If we don't get a literal expression, we're falling back and returning as is - so I added "try" to the name. Better yet would be to take @kotharironak 's suggestion and make sure it's a literal before passing in - then we could treat a non-literal as an error condition - easier to catch bugs.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/baseline/BaselineServiceImplTest.java
aaron-steinfeld: You could clone the mock for this spot if it's being shared in many places, but I suspect they'll mostly all want this change. We're changing internal logic, which is kind of the whole point of a unit test - the same data in and the same data out. The mocks would be the only thing to change, reflecting the new QS behavior. The fact that they're still coming through actually reflects lazy test writing - because both the arguments and results of the mock should change. Here, the arguments are listed as `any()`, which means the test isn't verifying that we're building the query we expect (which is a big part of what you're changing!)

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcherTests.java
sarthak77: resolved

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/QueryExpressionUtil.java
sarthak77: what about `convertLiteralExpressionToIsoFormat` ?

PR #112, opened 2021-12-10
fix: update log4j

aaron-steinfeld:
## Description
Upgrade vulnerable log4j

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,6 +5,6 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2023-31-31T00:00:00.000Z
+        expires: 2022-31-31T00:00:00.000Z
 
 patch: {}
--- a/.snyk
+++ b/.snyk
@@ -5,6 +5,6 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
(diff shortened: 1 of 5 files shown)


tim-mwangi (dismissed):

.snyk
```diff
@@ -5,6 +5,6 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2021-11-31T00:00:00.000Z
+        expires: 2023-31-31T00:00:00.000Z
```
tim-mwangi, line 8: lol. 2023-31?

aaron-steinfeld: oops, meant to do 2022!

gateway-service-api/build.gradle.kts
```diff
@@ -15,14 +15,14 @@ val generateLocalGoGrpcFiles = false
 
 protobuf {
   protoc {
-    artifact = "com.google.protobuf:protoc:3.17.3"
+    artifact = "com.google.protobuf:protoc:3.19.1"
```
aaron-steinfeld, line 18: So this introduces an incompatibility because we're using an older proto-java version. We should be declaring our proto java lib version (we get it transitively from grpc-protobuf today), but that will go viral quick so I'm not going to try to tackle that here.

rish691 (approved):

PR #116, opened 2021-12-30
Attribute expressions

aaron-steinfeld:
## Description
Replaced usages of column identifiers throughout with attribute expressions to support subpaths. Sorry for the large PR, difficult to add support piecemeal without breaking things. Requires QS release once https://github.com/hypertrace/query-service/pull/132 is reviewed and merged..

### Testing
Existing tests pass and verified E2E.


sarthak77:
Is spotless enabled for this ?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -284,37 +287,57 @@ public static Operator convertOperator(
...
+  private static AttributeExpression convertToQueryAttributeExpression(
       org.hypertrace.gateway.service.v1.common.ColumnIdentifier columnIdentifier) {
-    return ColumnIdentifier.newBuilder()
-        .setColumnName(columnIdentifier.getColumnName())
-        .setAlias(columnIdentifier.getAlias());
+    AttributeExpression.Builder builder =
+        AttributeExpression.newBuilder().setAttributeId(columnIdentifier.getColumnName());
+
+    Optional.of(columnIdentifier.getAlias())
+        .filter(Predicate.not(String::isEmpty))
```
kotharironak, line 310: Curious Q: Did we observe any issue as prior we were not checking for an empty string? If we are doing this here, do we need to check if not empty in the above function - `convertToQueryAttributeExpression`?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -471,8 +494,20 @@ private static boolean hasTimeRangeFilter(Filter filter, String timestampAttribu
...
           .anyMatch(f -> hasTimeRangeFilter(f, timestampAttributeId));
     }
-    return filter.getLhs().getValueCase() == Expression.ValueCase.COLUMNIDENTIFIER
-        && filter.getLhs().getColumnIdentifier().getColumnName().equals(timestampAttributeId);
+    return getSelectionAttributeId(filter.getLhs())
+        .map(filterAttributeId -> filterAttributeId.equals(timestampAttributeId))
+        .orElse(false);
+  }
+
+  private static Optional<String> getSelectionAttributeId(Expression expression) {
```
kotharironak, line 502: should we rename this `getSelectionAttributeId` -> `getAttributeId`? The first time, I was reading, I read something to do with the selection clause. What do you think?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
```diff
@@ -84,12 +85,8 @@ public EntityFetcherResponse getEntities(
     }
 
     // Add all expressions in the select that are already not part of the EntityID attributes
-    entitiesRequest.getSelectionList().stream()
-        .filter(expression -> expression.getValueCase() == ValueCase.COLUMNIDENTIFIER)
-        .filter(
-            expression ->
-                !mappedEntityIdAttributeIds.contains(
-                    expression.getColumnIdentifier().getColumnName()))
+    entitiesRequest
```
kotharironak, line 88: Curious Q: Why did we remove this filter `!mappedEntityIdAttributeIds`? Does this mean we will add those selections twice?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
```diff
@@ -212,4 +202,23 @@ public long getTotal(EntitiesRequestContext requestContext, EntitiesRequest enti
...
+    return request.getSelectionList().stream()
+        .filter(
+            expression ->
+                ExpressionReader.getSelectionAttributeId(expression)
+                    .map(attributeMetadataByIdMap::containsKey)
+                    .orElse(false))
+        .map(
+            expression ->
+                Map.entry(
+                    ExpressionReader.getSelectionResultName(expression).orElseThrow(),
```
kotharironak, line 219: Can we rename this as `ExpressionReader.getSelectionAttributeId` -> `ExpressionReader.getAttributeId`?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -539,37 +533,47 @@ private void parseResultSet(
...
+                Collectors.toUnmodifiableMap(
+                    expression ->
+                        ExpressionReader.getSelectionAttributeId(expression).orElseThrow(),
+                    expression ->
+                        ExpressionReader.getSelectionResultName(expression).orElseThrow()));
+
+    if (selectionResultNames.containsKey(FROM_ENTITY_ID_ATTRIBUTE_ID)) {
       interaction.putAttribute(
-          FROM_ENTITY_ID_ATTRIBUTE_ID,
+          selectionResultNames.get(FROM_ENTITY_ID_ATTRIBUTE_ID),
```
kotharironak, line 554: Prior, the alias was not used, was that taken care of some other place, or did this change fix the case of not handling alias?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
```diff
@@ -37,10 +32,10 @@ public RequestPreProcessor(
    * This is called once before processing the request.
    *
    * @param originalRequest The original request received
-   * @return The modified request with any additional filters depending on the scope config
+   * @return The modified request with unique selections and any additional filters depending on the
+   *     scope config
    */
-  public EntitiesRequest transformFilter(
-      EntitiesRequest originalRequest, EntitiesRequestContext context) {
+  public EntitiesRequest process(EntitiesRequest originalRequest, EntitiesRequestContext context) {
```
kotharironak, line 38: Q: Any specific reason to rename the method as we have one more down transformFilter for TraceRequest?  Are we planning to rename it as well in future?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/update/EdsEntityUpdater.java
```diff
@@ -134,4 +137,22 @@ private EntityUpdateRequest convertToEqsUpdateRequest(UpdateEntityRequest update
 
     return eqsUpdateRequestBuilder.build();
   }
+
+  private Map<String, AttributeMetadata> getAttributeMetadataByResultName(
+      UpdateEntityRequest updateRequest, UpdateExecutionContext updateExecutionContext) {
+    return updateRequest.getSelectionList().stream()
```
kotharironak, line 143: Looks like `AttributeMetadataUtil.remapAttributeMetadataByResultKey` can be reused here?

```
AttributeMetadataUtil.remapAttributeMetadataByResultKey(updateRequest.getSelectionList(), updateExecutionContext.getAttributeMetadata())
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -168,39 +171,39 @@ private ExploreRequest createRequest(
       ExploreRequest originalRequest, ExploreResponse.Builder originalResponse) {
     Filter.Builder filterBuilder = Filter.newBuilder();
     filterBuilder.setOperator(Operator.AND);
-    List<String> groupByColumns = groupByColumnList(originalRequest);
+    Map<String, Expression> groupBySelectionExpressionsByKey =
+        groupByExpressionByKey(originalRequest);
```
kotharironak, line 175: should we rename `groupByExpressionByKey` -> `groupByExpressionByResultKey` and `groupBySelectionExpressionsByKey` -> `groupBySelectionExpressionsByResultKey`?

kotharironak (dismissed):
overall, lgtm. Have only a few minor comments which can be taken care of as follow-up if we are doing.

It seems to me that the test for alias <resultname> changes are covered in the existing tests  - attribute having an alias, not having - both in JSON files and in EnityServiceTest asserts for map change. So, scan one more time that if we haven't missed something.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
aaron-steinfeld: On 295, we do this check in a more formal way via `hasAlias()`. This works because on the new GW AttributeExpression message, alias is optional and thus generates the presence check. On the existing column identifier message being used here, because it is not optional the only way to check if it's set is comparing it to the default value (empty string).

Prior to this change, when going from one col identifier to another, the check wasn't needed more or less for the same reason. The QS version of alias didn't support the presence check leaving no difference between an unset alias and one set to an empty string. Now, if we propagated the empty string col identifier alias into attribute expression, `.hasAlias()` would incorrectly return true.

A bit long winded, but does that make sense?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
aaron-steinfeld: What I was trying to convey here (and I think is missing from `getAttributeId`)  is that this only extracts attribute id from a col identifier or attribute expression, rather than an order by, function etc. I had taken to lumping the two of them together in naming as "selection", but I can see how that would be unclear. Any ideas that could better convey that distinction?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
aaron-steinfeld: The reason it was removed was that we were dropping alias information here - if the caller requested an id attribute by a different alias, the existing implementation was removing that. I had figured that duplicating a selection wouldn't have any significant impact, and wasn't worth the complexity.

That said, in the QS version of the entity fetcher, I believe I ended up with a better solution where I dropped dupe requests and used the aliases at the end to remap them back ( https://github.com/hypertrace/gateway-service/pull/116/files/ea0965d95e98c7caaa83dea5501dc17437362189#diff-11defa5d01d0ff51c801f8ea113e9b9100328b20e0d3073b3faae6bb38c31626R241-R246 ) so let me revisit this and see if that's feasible here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
aaron-steinfeld: Started discussion in https://github.com/hypertrace/gateway-service/pull/116/files/ea0965d95e98c7caaa83dea5501dc17437362189#r778870420

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
aaron-steinfeld: Correct, aliases were not working here before this change (well more accurately, graphql was using attribute IDs as aliases, so they matched).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
aaron-steinfeld: The reason I renamed it is this wasn't just transforming the filter, it was also modifying other parts of the request (selections), so the name "transformFilter" felt misleading. Since the class is called RequestPreProcessor, "process" seemed an apt name, but open to other suggestions.

In the meantime, will rename the trace version too.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
aaron-steinfeld: Sure, will use result name for consistency.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/update/EdsEntityUpdater.java
aaron-steinfeld: Thanks! I had implemented this multiple times before I realized it was identical and moved it to the shared util, must have missed fixing this one up.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityDataServiceEntityFetcher.java
aaron-steinfeld: Followed up with the same solution here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/transformer/RequestPreProcessor.java
aaron-steinfeld: Trace version renamed.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
aaron-steinfeld: renamed.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
kotharironak: Hmm!! Got it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
kotharironak: I see it's a bit tricky as the two names are very different 'columnIdentifier' and 'attributeExpression'. Not able to come up with a good name so that it is readable in the above context. Maybe any of the below (if not we can use the same name for now).

`<get|fetch|extract>AttributeIdFromColumnOrAttributeExpression` 
`getAttributeIdFromAttributeSelection` - As we are using, somewhere attributeSelection which refers to Column/AttributeExpression

aaron-steinfeld: I buy `getAttributeIdFromAttributeSelection` - will update to that.

aaron-steinfeld: done

kotharironak (approved):

PR #117, opened 2022-02-15
fix(entities): single source defaulting to EDS for inherent time filter

skjindal93:


avinashkolluru (dismissed):

gateway-service-baseline-lib/build.gradle.kts
```diff
@@ -15,8 +15,11 @@ dependencies {
   // Common utilities
   implementation("org.apache.commons:commons-lang3:3.10")
   implementation("org.apache.commons:commons-math:2.2")
-  implementation("com.google.protobuf:protobuf-java-util:3.17.3")
+  implementation("com.google.protobuf:protobuf-java-util:3.19.2")
   constraints {
+    implementation("com.google.protobuf:protobuf-java:3.19.2") {
```
aaron-steinfeld, line 20: We want to avoid constraints when possible and instead fix the source issues. Went through all the deps in https://github.com/hypertrace/gateway-service/pull/118 to separate those issues out of this PR.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -357,4 +356,54 @@ private QueryNode createQsDataFetcherNodeWithLimitAndOffset(EntitiesRequest enti
...
+     *   <li>single source is {@link EDS}
+     *   <li>inherent time filter is valid
+     *   <li>only live entities are requested (since non live entities request ignores the inherent
+     *       time filter anyways)
+     *   <li>timestamp column doesn't exist for the corresponding entity type
+     * </ul>
+     */
+    String singleSource = singleSourceForAllAttributes.get();
+    if (EDS.name().equals(singleSource)
+        && isValidTimeRange(
```
aaron-steinfeld, line 396: Do we support queries without a time range today? Doesn't seem like the spot to be doing validation

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -357,4 +356,54 @@ private QueryNode createQsDataFetcherNodeWithLimitAndOffset(EntitiesRequest enti
...
+            entitiesRequest.getStartTimeMillis(), entitiesRequest.getEndTimeMillis())
+        && !entitiesRequest.getIncludeNonLiveEntities()
+        && TimestampConfigs.getTimestampColumn(entitiesRequest.getEntityType()) == null) {
```
aaron-steinfeld, line 399: timestamp column always implies EDS?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -357,4 +356,54 @@ private QueryNode createQsDataFetcherNodeWithLimitAndOffset(EntitiesRequest enti
   private QueryNode createPaginateOnlyNode(QueryNode queryNode, EntitiesRequest entitiesRequest) {
     return new PaginateOnlyNode(queryNode, entitiesRequest.getLimit(), entitiesRequest.getOffset());
   }
+
+  private Optional<String> getSingleSourceForAllAttributes() {
+    // All expressions' attributes from the same source. Will only need one downstream query.
+    Optional<String> singleSourceForAllAttributes =
+        ExecutionTreeUtils.getSingleSourceForAllAttributes(executionContext);
```
aaron-steinfeld, line 363: We have two methods with the same name, but different behavior. Is the other one called in other places? Does this one belong in the utils file? If I were new to this code (which I am!) I wouldn't know which to use when.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: Can get rid of the validation, but seemed like a valid validation, since it's moved under utils now

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: yes

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: Moved into a single method inside the utils

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
aaron-steinfeld: it's not that it's not valid - it's just not the purpose of this method - single responsibility!

PR #121, opened 2022-03-17
feat: parallelize span, trace total and event fetches

aaron-steinfeld:
## Description
Improve performance of trace and span queries by fetching the result rows and total in parallel.

### Testing
Reverified unit, e2e tests

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
@@ -7,6 +7,7 @@ import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.StreamObserver;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
 import org.apache.commons.lang3.StringUtils;
 import org.hypertrace.core.attribute.service.client.AttributeServiceClient;
 import org.hypertrace.core.attribute.service.client.config.AttributeServiceClientConfig;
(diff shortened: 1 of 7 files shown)


gateway-service-impl/src/test/resources/configs/gateway-service/application.conf
```diff
@@ -17,6 +17,9 @@ attributes.service.config = {
   host = localhost
   port = 9012
 }
+query.executor.config = {
```
kotharironak, line 20: Can we add the helm too?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
```diff
@@ -79,6 +82,8 @@ public GatewayServiceImpl(Config appConfig) {
     QueryServiceClient queryServiceClient =
         new QueryServiceClient(new QueryServiceConfig(qsConfig));
     int qsRequestTimeout = getRequestTimeoutMillis(qsConfig);
+    ExecutorService queryExecutor =
+        QueryExecutorServiceFactory.buildExecutorService(QueryExecutorConfig.from(appConfig));
```
kotharironak, line 86: are we use `ServiceFactory` suffix other places? if, not, `QueryExecutorServiceFactory` -> `QueryExecutorFactory` ?

gateway-service-impl/src/test/resources/configs/gateway-service/application.conf
kotharironak: can be added in follow-up.

kotharironak (approved):
lgtm

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/GatewayServiceImpl.java
aaron-steinfeld: So I went back and forth on this - agreed we should be consistent, but the actual returned interface is an ExecutorService, so it's a factory for that, and the role of that executor service is querying.

Will proceed, but let's figure out the naming and can go back and make consistent.

gateway-service-impl/src/test/resources/configs/gateway-service/application.conf
aaron-steinfeld: the helm config map template takes in a blob for application conf, so no changes are needed there (I could have overridden this default, but that didn't seem necessary):

https://github.com/hypertrace/gateway-service/blob/main/helm/values.yaml#L67-L76

PR #122, opened 2022-03-19
Parallelize trace queries

aaron-steinfeld:
## Description
Limit total agg queries that only expect a single result. Allows better query validation on QS side.


aaron-steinfeld:
Whoopps wrong branch. recreating.

PR #124, opened 2022-03-20
refactor: reduce grouped time series explorer limit

aaron-steinfeld:
## Description
Fixed two issues in explorer
1. When processing results, metadata lookup was failing as it was assuming the result col name must be the attribute id, when it's actually the provided alias
2. Rather than over-limiting on TS queries, we limit to exactly the amount of returned results. I want to verify this is right, but my thinking is this is holdover from before we had separate group and row limits.

### Testing
Updated and verified UTs, tested behavior e2e
### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
@@ -6,7 +6,9 @@ import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.query.service.api.ColumnMetadata;
 import org.hypertrace.core.query.service.api.QueryRequest;
 import org.hypertrace.core.query.service.api.ResultSetMetadata;
@@ -63,12 +65,7 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
     addGroupByExpressions(builder, request);
 
     // 5. Set Limit.
-    // Scale the limit size based on the limit so that we have a better chance of capturing all the
-    // results within the
-    // time range. This is especially important when the actual Group By list is not empty.
-    builder.setLimit(
-        Math.min(request.getLimit(), 1000)
-            * QueryServiceClient.DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+    builder.setLimit(request.getLimit());
     requestContext.setOrderByExpressions(getRequestOrderByExpressions(request));
 
     return builder.build();
@@ -182,7 +179,7 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
       ResultSetMetadata resultSetMetadata,
       ExploreResponse.Builder builder,
       ExploreRequestContext requestContext,
-      AttributeMetadataProvider attributeMetadataProvider) {
+      Map<String, AttributeMetadata> resultKeyToAttributeMetadataMap) {
     var rowBuilder = org.hypertrace.gateway.service.v1.common.Row.newBuilder();
 
     // First column is the time column. (Also the column name is "dateTimeConvert", Pinot's function
@@ -203,7 +200,7 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
           resultSetMetadata.getColumnMetadata(i),
           rowBuilder,
           requestContext,
-          attributeMetadataProvider);
+          resultKeyToAttributeMetadataMap);
     }
     builder.addRow(rowBuilder);
   }
@@ -214,7 +211,7 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
       ColumnMetadata metadata,
       org.hypertrace.gateway.service.v1.common.Row.Builder rowBuilder,
       ExploreRequestContext requestContext,
-      AttributeMetadataProvider attributeMetadataProvider) {
+      Map<String, AttributeMetadata> resultKeyToAttributeMetadataMap) {
     TimeAggregation timeAggregation =
         requestContext.getTimeAggregationByAlias(metadata.getColumnName());
     if (timeAggregation != null) { // Time aggregation with Function expression value
@@ -222,12 +219,11 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
           queryServiceValue,
           metadata,
           rowBuilder,
-          requestContext,
-          attributeMetadataProvider,
+          resultKeyToAttributeMetadataMap,
           timeAggregation.getAggregation().getFunction());
     } else { // Simple columnId Expression value eg. groupBy columns or column selections
       handleQueryServiceResponseSingleColumn(
-          queryServiceValue, metadata, rowBuilder, requestContext, attributeMetadataProvider, null);
+          queryServiceValue, metadata, rowBuilder, resultKeyToAttributeMetadataMap, null);
     }
   }
 
--- a/gateway-service-impl/build.gradle.kts
+++ b/gateway-service-impl/build.gradle.kts
@@ -28,7 +28,7 @@ dependencies {
   implementation("com.google.guava:guava:30.1.1-jre")
 
   implementation("com.fasterxml.jackson.core:jackson-annotations:2.13.2")
-  implementation("com.fasterxml.jackson.core:jackson-databind:2.13.2")
+  implementation("com.fasterxml.jackson.core:jackson-databind:2.13.2.2")
 
   testImplementation("org.junit.jupiter:junit-jupiter:5.8.2")
   testImplementation("org.mockito:mockito-core:4.3.1")
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
@@ -5,6 +5,9 @@ import com.google.protobuf.util.JsonFormat;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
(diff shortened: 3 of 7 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
```diff
@@ -63,12 +65,7 @@ QueryRequest buildQueryRequest(
     addGroupByExpressions(builder, request);
 
     // 5. Set Limit.
-    // Scale the limit size based on the limit so that we have a better chance of capturing all the
-    // results within the
-    // time range. This is especially important when the actual Group By list is not empty.
-    builder.setLimit(
```
aaron-steinfeld, line 69: @tim-mwangi do you have any recollection on why we did this? The explanation doesn't make sense to me.

In the ungrouped TS case, the limit is whatever was provided - if it cuts off intervals, that's the right behavior.

In the grouped TS case, same thing - and I think maybe this changed since the initial impl (before group limit was added). If I have a limit 6 and ask for 3 groups, I'd expect 6 results back, not 18. Given that, does this change seem correct to you?

tim-mwangi: Back then it seems like whenever we had group by we always did the sorting, paging and offsetting ourselves hence the need to cast a very wide net. See https://github.com/hypertrace/gateway-service/blob/2255fd7c49a42a2f81fccaee1b8574feb4136cd5/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java#L65 and that's the function which we override here. It seems like we treat time aggregations like a group by request hence similar logic in here. 

If Pinot has implemented sorting by the group column or aggregation column when group by is specified, then we can get rid of this. I remember it had to do with lack of sorting support in Pinot when it's a group by query. I don't recall the conditions for the lack of support though.

aaron-steinfeld: > If Pinot has implemented sorting by the group column or aggregation column when group by is specified, then we can get rid of this. I remember it had to do with lack of sorting support in Pinot when it's a group by query. I don't recall the conditions for the lack of support though.

I'm not sure if they have, but we also don't sort time series do we? 

If it's an ungrouped TS, I'm assuming pinot sorts by tie, but whatever it's default it seems acceptable - the requested limit is less than the requested intervals, so it's kind of a nonsensical query.

If it's grouped, we have the behavior you indicated on the first group by request, but that's all before we get here. We then have the same logic as above - a known number of groups and a user limit, with no sorting to worry about.

Does that seem accurate? If this is just trading some potential noondeterminism on nonsense queries for performance and appropriate limits on expected queries, that seems a good tradeoff.

tim-mwangi: Aah here it is https://github.com/hypertrace/gateway-service/blob/2255fd7c49a42a2f81fccaee1b8574feb4136cd5/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandlerWithSorting.java#L8.
```
/**
 * Extension of IRequestHandler that will also implement methods to manually sort results from the
 * query service. This is needed when Group By is in the request as Pinot does not support Group By
 * with Order By.
 */
```

aaron-steinfeld: We don't expose or support ordering by a group by request as far as I'm aware - this sounds like we went above and beyond to create in memory support for a feature that's not exposed.

aaron-steinfeld: Ahh maybe we do, just not on the explorer screen itself. Need to dig into that a bit. We definitely don't use it with time series though.

aaron-steinfeld: So @tim-mwangi assuming we don't need support for a sorted TS, are we good to go here?

tim-mwangi: Let me ask this: How do non-explorer queries handle time-series? https://github.com/hypertrace/gateway-service/blob/ed5b7bbf8eba06116beadeed1c1940aed8a52463/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/QueryServiceEntityFetcher.java#L617 . It seems like we also set a huge limit here. I don't know if anything has changed in Pinot that would allow us to change this. Are sending the correct TS limit for both scenarios from the client? Also even if we are doing this it is not guaranteed to be sorted by Pinot by the TS column(last I checked)

Also there's TimeAggregationsWithGroupBy to consider https://github.com/hypertrace/gateway-service/blob/ed5b7bbf8eba06116beadeed1c1940aed8a52463/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java#L20

aaron-steinfeld: >  It seems like we also set a huge limit here. I don't know if anything has changed in Pinot that would allow us to change this. Are sending the correct TS limit for both scenarios from the client? 

The difference is revealed by that comment there. In the entity service time series case, the request's limit refers to the number of entities to return - there's no specification of the number of time series intervals to return. This stands in contrast to the explorer request where two different limits are set - the group limit, which should be the limit when sending a group by request, and the plain old limit, which is the total number of result rows. For an aggregation group by request without a rest group, specifying either should result in the same behavior (rest group just would mean one less group). For a non grouped time series, only the row limit is relevant so we don't need to overfetch to satisfy that (unless we need to in memory sort, which I'll get to in a sec). For a grouped time series, the exact same logic applies, since the groups have already been limited and that results in the appropriate filter.

So this brings us back to the sorting - if I ask for 1h intervals over the last day with a limit of 5 rows, do I expect the first 5? last 5? random 5? I think we could define some reasonable expectations on this, but I do not think this is an important enough use case to force ourselves to do in memory sorting to satisfy those expectations given that the result count is a product of the same request. The user has given us conflicting asks, and can fix that to explicitly define what data they want (e.g. retaining the limit 5 and reducing the time range to the first 5 or last 5 hours; decreasing any group limit; increasing the interval size etc.)

Does that make sense? Do you see any use cases outside of that scenario (or things I missed within) that would be affected by this change or disagree with this line of thinking?

aaron-steinfeld: FWIW - I went to pinot and tested this with a 24h query at 1h interval and a 6 result limit - I don't know how far it extrapolates, but I saw the following behavior:

No order by - random results - not in sequence nor consecutive.
```
Select dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS'), COUNT(*) FROM spanEventView WHERE customer_id = '<tenantId>' AND  start_time_millis >= 1647892950000 AND start_time_millis < 1647979354000 GROUP BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS') limit 6
```

Order by time (default ascending) - starting from 24h ago:
```
Select dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS'), COUNT(*) FROM spanEventView WHERE customer_id = '<tenantId>' AND  start_time_millis >= 1647892950000 AND start_time_millis < 1647979354000 GROUP BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS') ORDER BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS') limit 6
```
Order by time DESC - starting from now:
```
Select dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS'), COUNT(*) FROM spanEventView WHERE customer_id = '<tenantId>' AND  start_time_millis >= 1647892950000 AND start_time_millis < 1647979354000 GROUP BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS') ORDER BY dateTimeConvert(start_time_millis,'1:MILLISECONDS:EPOCH','1:MILLISECONDS:EPOCH','3600:SECONDS') DESC limit 6
```

tim-mwangi: Huh. So they are doing order by when group by is specified now. In that case we can stop doing the in memory sorting and offsetting.

skjindal93 (approved):

aaron-steinfeld:
Happy to pull out the attribute stuff and rely on https://github.com/hypertrace/gateway-service/pull/126 - will try to update tonight or tomorrow.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
aaron-steinfeld: Yeah, agree there's big improvements that can be done there but not sure they're in scope for this particular PR - here, we're still doing in memory sorting, we're just potentially dropping results out of order if and only if a user has a contradictory request. Unless you have a concern with that compromise or see another issue, let's put off the sorting improvements for another day.

tim-mwangi: ok. sounds good.

aaron-steinfeld:
@hypertrace/query-layer-owners bump

rish691 (approved):

PR #127, opened 2022-04-04
fix(explore): order by on eds attributes

skjindal93:
## Description
Order by on EDS attributes in explore, had missing order by expressions set on the explore request context, which are needed later to group by, and order in memory later

### Testing
Verified the changes end to end

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
@@ -4,6 +4,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.hypertrace.core.query.service.client.QueryServiceClient;
 import org.hypertrace.entity.query.service.client.EntityQueryServiceClient;
 import org.hypertrace.entity.query.service.v1.EntityQueryRequest;
 import org.hypertrace.entity.query.service.v1.Expression;
@@ -17,8 +18,10 @@ import org.hypertrace.gateway.service.common.AttributeMetadataProvider;
 import org.hypertrace.gateway.service.common.converters.EntityServiceAndGatewayServiceConverter;
 import org.hypertrace.gateway.service.common.util.AttributeMetadataUtil;
 import org.hypertrace.gateway.service.common.util.ExpressionReader;
+import org.hypertrace.gateway.service.common.util.OrderByUtil;
 import org.hypertrace.gateway.service.entity.config.EntityIdColumnsConfigs;
 import org.hypertrace.gateway.service.explore.ExploreRequestContext;
+import org.hypertrace.gateway.service.v1.common.OrderByExpression;
 import org.hypertrace.gateway.service.v1.explore.ExploreRequest;
 
 public class EntityServiceEntityFetcher {
@@ -55,6 +58,18 @@ public class EntityServiceEntityFetcher {
 
     addGroupBys(exploreRequest, builder);
     addSelections(requestContext, exploreRequest, builder);
+
+    // TODO: Push group by down to EQS
+    // If there is a group by, specify a large limit and track actual limit, offset and order by
+    // expression list, so we can compute these once the we get the results.
+    if (requestContext.hasGroupBy()) {
+      // Will need to do the ordering, limit and offset ourselves after we get the group by results
+      builder.setLimit(QueryServiceClient.DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+      requestContext.setOrderByExpressions(getRequestOrderByExpressions(exploreRequest));
+    } else {
+      // No Group By
+      throw new RuntimeException("Entity request handler only supports group by requests");
+    }
 
     return builder.build();
   }
@@ -118,5 +133,10 @@ public class EntityServiceEntityFetcher {
             EntityServiceAndGatewayServiceConverter.convertToEntityServiceFilter(
                 exploreRequest.getFilter()))
         .addAllChildFilter(entityIdsInFilter);
+  }
+
+  private List<OrderByExpression> getRequestOrderByExpressions(ExploreRequest request) {
+    return OrderByUtil.matchOrderByExpressionsAliasToSelectionAlias(
+        request.getOrderByList(), request.getSelectionList(), request.getTimeAggregationList());
   }
 }
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
@@ -99,6 +99,7 @@ public class RequestHandler implements RequestHandlerWithSorting {
     // 3. Add GroupBy
     addGroupByExpressions(builder, request);
 
+    // TODO: Push group by down to QS
     // 4. If there's no Group By, Set Limit, Offset and Order By.
     // Otherwise, specify a large limit and track actual limit, offset and order by expression list
     // so we can compute


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -56,6 +59,18 @@ private EntityQueryRequest buildRequest(
     addGroupBys(exploreRequest, builder);
     addSelections(requestContext, exploreRequest, builder);
 
+    // TODO: Push group by down to EQS
+    // If there is a group by, specify a large limit and track actual limit, offset and order by
+    // expression list, so we can compute these once the we get the results.
+    if (requestContext.hasGroupBy()) {
+      // Will need to do the ordering, limit and offset ourselves after we get the group by results
+      builder.setLimit(QueryServiceClient.DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
```
aaron-steinfeld, line 67: nit - this limit is going to ES but is the QS default. Should be renamed or specified separately for ES.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -56,6 +59,18 @@ private EntityQueryRequest buildRequest(
...
+      requestContext.setOrderByExpressions(getRequestOrderByExpressions(exploreRequest));
+    } else {
+      // No Group By
+      throw new RuntimeException("Entity request handler only supports group by requests");
```
aaron-steinfeld, line 71: Why? What would fail in this path (other than us not assigning the requested limit and order by)?

aaron-steinfeld, line 71: Why? What would fail in this path (other than us not assigning the requested limit and order by)?

skjindal93: Currently, the assumption is entity handler will only be used, in case there is a group by (and filters, order bys, and selections) on EDS attributes.

Since, we return a row for every span (_and not a row per entity type_), we can think about extending the support, where we get the attribute value from EDS per entity, and then copy the value across all the rows for the same entity, but there isn't a use case yet

And, I am actually thinking of changing the implementation to return a single row per entity, if the explore context is an entity type, and not return a row for every span

aaron-steinfeld: Another tidbit here though, if I ask a query like "give me the count of entity ids grouped by some QS attribute", I would expect that count of entity IDs to be based on entity records, not spans. Seems like things are getting fun 😉

skjindal93: > I would prioritize aggregation support (with or without group by) over individual record querying.

Aggregation support without group by? What does that mean? 🤔

skjindal93: Added support for the above

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -57,7 +60,18 @@ private EntityQueryRequest buildRequest(
...
     addSelections(requestContext, exploreRequest, builder);
+
     builder.setLimit(DEFAULT_ENTITY_REQUEST_LIMIT);
+
+    // TODO: Push group by down to EQS
+    // If there is a group by, specify a large limit and track actual limit, offset and order by
+    // expression list, so we can compute these once the we get the results.
+    if (requestContext.hasGroupBy()) {
+      // Will need to do the ordering, limit and offset ourselves after we get the group by results
+      builder.setLimit(DEFAULT_ENTITY_REQUEST_LIMIT);
```
aaron-steinfeld, line 71: How does this differ from line 64?

aaron-steinfeld, line 71: How does this differ from line 64?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -57,7 +60,18 @@ private EntityQueryRequest buildRequest(
...
     addGroupBys(exploreRequest, builder);
     addSelections(requestContext, exploreRequest, builder);
+
     builder.setLimit(DEFAULT_ENTITY_REQUEST_LIMIT);
+
+    // TODO: Push group by down to EQS
+    // If there is a group by, specify a large limit and track actual limit, offset and order by
+    // expression list, so we can compute these once the we get the results.
+    if (requestContext.hasGroupBy()) {
+      // Will need to do the ordering, limit and offset ourselves after we get the group by results
```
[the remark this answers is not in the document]
skjindal93: We aren't pushing order by to EDS in case of explore. EDS handler is only called when 
- there is a group by, order by and selections on EDS. In this case, we handle it in memory. We can push group by to EDS along with order by, which is being called out by the TODO comment. That requires a separate effort
- there is an aggregated selection on EDS (like DISTINCT COUNT of APIs). There is no order by here. And, the limit in this case would be 1 inherently

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
skjindal93: fixed it.

gateway-service-factory/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-  api("org.hypertrace.core.serviceframework:platform-grpc-service-framework:0.1.58")
+  api("org.hypertrace.core.serviceframework:platform-grpc-service-framework:0.1.59")
```
aaron-steinfeld, line 6: can we bring this up to date with latest? Also guessing there's more framework libs that should be brought up to same version - at least metrics should be hiding somewhere, since I know we have them for GW.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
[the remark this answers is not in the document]
aaron-steinfeld: Second case makes sense.

First case - I thought we _were_ pushing down group by (line 61). Discussed offline, but since this is the final query and isn't being joined with anything in memory (right?) we should be able to push everything down unless there's something missing support in EQS.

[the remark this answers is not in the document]
skjindal93: Tried out different queries. We don't support ordering in document store on function expressions. Since, that's the case, we can't push limit/offset and order bys to EQS

We are already pushing group bys to EQS. Pushing limit and offset to EQS, only if there are no order by expressions (which shouldn't ideally be the case, since group by mostly comes with order by)

```
orderBy: [
  {
    aggregation: COUNT
    size: null
    direction: DESC
    keyExpression: { key: "name" }
  }
]
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -60,16 +63,19 @@ private EntityQueryRequest buildRequest(
...
-    // expression list, so we can compute these once the we get the results.
-    if (requestContext.hasGroupBy()) {
+    addLimitAndOffset(requestContext, exploreRequest, builder);
+
+    // TODO: Push order by down to EQS
+    // EQS (and document-store) currently doesn't support order by on functional expressions
+    // If there are order by expressions, specify a large limit and track actual limit, offset and
+    // order by
+    // expression list, so we can compute these once we get the results.
+    if (!requestContext.getOrderByExpressions().isEmpty()) {
```
aaron-steinfeld, line 73: Can we move this code into addLimitAndOffset? It's confusing that we set the limit and offset unconditionally (in addLimitAndOffset), then conditionally overwrite them in the top level buildRequest call.

Also whether we push down or not will influence response processing. We should make sure we're in sync (i.e. have common code or state so we only evaluate the condition in one place in code).

PR #129, opened 2022-04-26
fix(explore): subpath queries

skjindal93:
## Description
Subpath attribute expressions were picking up the same value type/attribute kind as map, instead of primitive value types inside the map

Basically, if the query says 

```
url: selection(
  expression: { key: "spanTags", subpath: "url" }
) {
  value
}
```
the attribute kind for `url` becomes `TYPE_STRING_MAP`, since the attribute kind for `spanTags` is `TYPE_STRING_MAP`. This was leading to "Unable to read Map JSON String data" error, since the value for the subpath `url` isn't really a map

### Testing
Tested end to end

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/QueryRequestContext.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/QueryRequestContext.java
@@ -3,12 +3,14 @@ package org.hypertrace.gateway.service.common;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import org.hypertrace.gateway.service.v1.common.AttributeExpression;
 import org.hypertrace.gateway.service.v1.common.FunctionExpression;
 import org.hypertrace.gateway.service.v1.common.TimeAggregation;
 
 // Hold some of request fields and mappings needed when parsing the query service response
 public class QueryRequestContext extends RequestContext {
   private final Map<String, FunctionExpression> aliasToFunctionExpressionMap = new HashMap<>();
+  private final Map<String, AttributeExpression> aliasToMapAttributeExpressionMap = new HashMap<>();
   private final long startTimeMillis;
   private final long endTimeMillis;
   private final Map<String, TimeAggregation> aliasToTimeAggregation = new HashMap<>();
@@ -30,6 +32,15 @@ public class QueryRequestContext extends RequestContext {
 
   public FunctionExpression getFunctionExpressionByAlias(String alias) {
     return aliasToFunctionExpressionMap.get(alias);
+  }
+
+  public void mapAliasToMapAttributeExpression(
+      String alias, AttributeExpression mapAttributeExpression) {
+    aliasToMapAttributeExpressionMap.put(alias, mapAttributeExpression);
+  }
+
+  public AttributeExpression getMapAttributeExpressionByAlias(String alias) {
+    return aliasToMapAttributeExpressionMap.get(alias);
   }
 
   public long getStartTimeMillis() {
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
@@ -572,7 +572,7 @@ public class QueryAndGatewayDtoConverter {
   }
 
   public static org.hypertrace.gateway.service.v1.common.Value
-  convertToGatewayValueForMapAttributeValue(
+      convertToGatewayValueForMapAttributeValue(
           AttributeKind attributeKind,
           Map<String, AttributeMetadata> resultKeyToAttributeMetadataMap,
           ColumnMetadata metadata,
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
@@ -570,4 +570,41 @@ public class QueryAndGatewayDtoConverter {
     return QueryAndGatewayDtoConverter.convertQueryValueToGatewayValue(
         queryValue, attributeMetadata);
   }
+
+  public static org.hypertrace.gateway.service.v1.common.Value
+  convertToGatewayValueForMapAttributeValue(
+          AttributeKind attributeKind,
+          Map<String, AttributeMetadata> resultKeyToAttributeMetadataMap,
+          ColumnMetadata metadata,
+          Value queryValue) {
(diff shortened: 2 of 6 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/QueryRequestContext.java
```diff
@@ -32,6 +34,15 @@ public FunctionExpression getFunctionExpressionByAlias(String alias) {
     return aliasToFunctionExpressionMap.get(alias);
   }
 
+  public void mapAliasToMapAttributeExpression(
```
aaron-steinfeld, line 37: this "request context" is a real monster huh

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -286,7 +310,16 @@ void handleQueryServiceResponseSingleColumn(
               resultKeyToAttributeMetadataMap,
               metadata,
               queryServiceValue);
-    } else { // Simple columnId Expression value eg. groupBy columns or column selections
+    } else if (mapAttributeExpression != null) {
```
aaron-steinfeld, line 313: if function != null and mapAttribute != null this wouldn't be hit, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -99,6 +104,21 @@ QueryRequest buildQueryRequest(
...
     addGroupByExpressions(builder, request);
 
+    columnSelections.stream()
+        .filter(not(ExpressionReader::isSimpleAttributeSelection))
+        .forEach(
+            expression ->
+                requestContext.mapAliasToMapAttributeExpression(
+                    expression.getAttributeExpression().getAlias(),
+                    expression.getAttributeExpression()));
+    request.getGroupByList().stream()
```
aaron-steinfeld, line 114: Do we care about group by if we're not selecting it back? Can't recall if that's auto-fetched or something

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -570,4 +570,41 @@ public static org.hypertrace.gateway.service.v1.common.Value convertToGatewayVal
...
+          Value queryValue) {
+    String resultKey = metadata.getColumnName();
+    // if there's no override from the map, then this is regular attribute, then use the attribute
+    // map to convert the value
+    AttributeMetadata attributeMetadata;
+    if (attributeKind == null) {
+      attributeMetadata = resultKeyToAttributeMetadataMap.get(resultKey);
+    } else {
+      switch (attributeKind) {
+        case TYPE_STRING_MAP:
```
aaron-steinfeld, line 588: So after all this, we assume it would be a string - which makes sense - but do we really need all that complexity given that assumption? It seems like we really only need to accept the attribute kind and default to string for when we can't look it up.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
skjindal93: The assumption is group bys are included as part of selections. Look at the comment here https://github.com/hypertrace/gateway-service/blob/b505531c68f211fa9ed3dbbeda75daa3124d4479/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java#L175

```
// The expectation from the Gateway service client is that they
// do not add the
// group by expressions to the selection expressions in the request
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
skjindal93: For a single column, both of them can't be non-null. The function handles response for a single column.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/QueryRequestContext.java
skjindal93: oh yes

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
skjindal93: I tried to design it a bit generically to support additional maps in future.

Which complexity though? Passing around the attribute expression, or mapping the alias to attribute map expression?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
aaron-steinfeld: we can't call a function on a map attribute? e.g. `distinct_count(tags.span.type)`

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
aaron-steinfeld: ah, strange. gotcha

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
aaron-steinfeld: both - I would think that we could just pass in the type, and when we lookup, if we can't map an attribute expression to  metadata (or more specifically, it has a subpath and the key maps to a map type) we just can default to string. That doesn't seem like it requires holding state in the request context or anything.

For other types of maps, we could add similar logic - e.g number map + subpath => number

skjindal93: I didn't quite follow. How would you know it's an attribute expression, without passing in the attribute expression around?

And, in order to get handle of the attribute expression, you would need to map the alias to attribute expression in request context, while building the request. And, then get the attribute expression from request context, while building the response

aaron-steinfeld: discussed this with @skjindal93 offline and concluded on an alternative approach to attempt to avoid holding state in context and generally reduce complexity - update the conversion methods to work with an AttributeKind (which is all they read off of metadata, anyway), rather than an AttributeMetadata. This will allow all types, map or otherwise, to work consistently and be resolved at the same time/place.

PR #130, opened 2022-05-04
fix(explore): subpath queries

skjindal93:
## Description
Subpath attribute expressions were picking up the same value type/attribute kind as map, instead of primitive value types inside the map

Basically, if the query says 

```
url: selection(
  expression: { key: "spanTags", subpath: "url" }
) {
  value
}
```
the attribute kind for `url` becomes `TYPE_STRING_MAP`, since the attribute kind for `spanTags` is `TYPE_STRING_MAP`. This was leading to "Unable to read Map JSON String data" error, since the value for the subpath `url` isn't really a map

### Testing
Tested end to end

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
@@ -7,6 +7,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.hypertrace.core.attribute.service.v1.AttributeKind;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.attribute.service.v1.AttributeScope;
 import org.hypertrace.gateway.service.common.AttributeMetadataProvider;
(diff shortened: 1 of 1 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/AttributeMetadataUtil.java
```diff
@@ -128,9 +119,32 @@ public static Map<String, AttributeMetadata> remapAttributeMetadataByResultKey(
...
+    if (maybeAttributeMetadata.isEmpty()) {
+      return Optional.empty();
+    }
+
+    AttributeMetadata attributeMetadata = maybeAttributeMetadata.get();
+    if (!ExpressionReader.isSimpleAttributeSelection(expression)) {
+      switch (attributeMetadata.getValueKind()) {
+        case TYPE_STRING_MAP:
+          return Optional.of(
+              AttributeMetadata.newBuilder(attributeMetadata)
```
aaron-steinfeld, line 142: I don't love creating a fake attribute metadata just to hold kind, but I do like this solution better in general. What was the issue you ran in when working with kinds rather than the general metadata object?

skjindal93: The issue was mostly the change size. It was leading to lots of refactor, and tests. Given, that it's a P1, I came up with a short workaround right now for this by generating fake metadata (_since we already follow that approach somewhere else_)

I can take the refactor up as a part of separate PR

aaron-steinfeld (approved):

PR #131, opened 2022-05-12
Support for MAX, MIN, SUM on timestamp 

japoorv:
Adding support for MAX, MIN, SUM on timestamp

### DIFF
--- a/gateway-service-impl/build.gradle.kts
+++ b/gateway-service-impl/build.gradle.kts
@@ -35,4 +35,8 @@ dependencies {
   testImplementation("org.mockito:mockito-inline:4.3.1")
   testImplementation("org.apache.logging.log4j:log4j-slf4j-impl:2.17.1")
   testImplementation("io.grpc:grpc-netty:1.44.0")
+
+  constraints {
+    testRuntimeOnly("io.netty:netty-common:4.1.77.Final")
+  }
 }
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
@@ -70,7 +70,7 @@ public class StringToAttributeKindConverter extends ToAttributeKindConverter<Str
 
       case TYPE_TIMESTAMP:
         valueBuilder.setValueType(ValueType.TIMESTAMP);
-        valueBuilder.setTimestamp(Long.parseLong(value));
+        valueBuilder.setTimestamp(Double.valueOf(value).longValue());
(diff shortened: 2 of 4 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -70,7 +70,7 @@ public Value doConvert(String value, AttributeKind attributeKind, Value.Builder
 
       case TYPE_TIMESTAMP:
         valueBuilder.setValueType(ValueType.TIMESTAMP);
-        valueBuilder.setTimestamp(Long.parseLong(value));
+        valueBuilder.setTimestamp(Double.valueOf(value).longValue());
```
avinashkolluru, line 73: what is the reason for changing this?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/MetricAggregationFunctionUtil.java
```diff
@@ -76,7 +76,8 @@ public static AttributeKind getValueTypeForFunctionType(
         // Min/Max/Sum function only applicable to numerical data
         Preconditions.checkArgument(
```
avinashkolluru, line 77: Not sure if SUM makes sense for TIMESTAMP

japoorv: Okay

avinashkolluru (approved):

gateway-service-impl/build.gradle.kts
```diff
@@ -35,4 +35,8 @@ dependencies {
   testImplementation("org.mockito:mockito-inline:4.3.1")
   testImplementation("org.apache.logging.log4j:log4j-slf4j-impl:2.17.1")
   testImplementation("io.grpc:grpc-netty:1.44.0")
+
+  constraints {
+    testRuntimeOnly("io.netty:netty-common:4.1.77.Final")
```
aaron-steinfeld, line 40: unnecessary - we don't enforce runtime vulnerabilities on test dependencies

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
aaron-steinfeld: What's being stored in scientific notation? That sounds like a bug somewhere - especially because this hasn't historically been an issue. If we expect to represent it as a long, it should be stored such, since we're losing precision here.

PR #132, opened 2022-05-24
if no time range set use entity request handler

d-trace:
## Description
Make `explore` query time range params optional

### Testing
Ran modified version of service with graphql service and ran the query

### Checklist:
- [ x ] My changes generate no new warnings

### Documentation

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
@@ -129,7 +129,7 @@ public class ExploreService {
               request.getGroupByList());
       Optional<String> source =
           ExpressionContext.getSingleSourceForAllAttributes(expressionContext);
-      if (source.isPresent() && EDS.toString().equals(source.get())) {
+      if ((source.isPresent() && EDS.toString().equals(source.get())) || !timeRangeSet(request)) {
         return entityRequestHandler;
       }
     }
@@ -140,6 +140,7 @@ public class ExploreService {
     if (hasTimeAggregations(request)) {
       return timeAggregationsRequestHandler;
     }
+
     return normalRequestHandler;
   }
 
@@ -149,5 +150,9 @@ public class ExploreService {
 
   private boolean hasTimeAggregationsAndGroupBy(ExploreRequest request) {
     return !request.getTimeAggregationList().isEmpty() && !request.getGroupByList().isEmpty();
+  }
+
+  private boolean timeRangeSet(ExploreRequest request) {
+    return request.getStartTimeMillis() != 0L && request.getEndTimeMillis() != 0;
   }
 }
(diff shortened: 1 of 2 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -150,4 +151,8 @@ private boolean hasTimeAggregations(ExploreRequest request) {
   private boolean hasTimeAggregationsAndGroupBy(ExploreRequest request) {
     return !request.getTimeAggregationList().isEmpty() && !request.getGroupByList().isEmpty();
   }
+
+  private boolean timeRangeSet(ExploreRequest request) {
+    return request.getStartTimeMillis() != 0L && request.getEndTimeMillis() != 0L;
```
skjindal93, line 156: Can we instead make the time range field optional in GRPC? Introducing optional shouldn't be a breaking change

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -150,4 +151,8 @@ private boolean hasTimeAggregations(ExploreRequest request) {
   private boolean hasTimeAggregationsAndGroupBy(ExploreRequest request) {
     return !request.getTimeAggregationList().isEmpty() && !request.getGroupByList().isEmpty();
   }
+
+  private boolean timeRangeSet(ExploreRequest request) {
```
skjindal93, line 155:
```suggestion
  private boolean hasTimeRange(ExploreRequest request) {
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -129,7 +129,7 @@ private IRequestHandler getRequestHandler(
               request.getGroupByList());
       Optional<String> source =
           ExpressionContext.getSingleSourceForAllAttributes(expressionContext);
-      if (source.isPresent() && EDS.toString().equals(source.get())) {
+      if ((source.isPresent() && EDS.toString().equals(source.get())) || !timeRangeSet(request)) {
```
skjindal93, line 132: We should error out in case, the time range isn't set and source is not EDS

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
d-trace: All fields are optional in proto3. Do you mean introducing oneOf instead https://stackoverflow.com/questions/42622015/how-to-define-an-optional-field-in-protobuf-3?

d-trace: Done

skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -129,9 +131,12 @@ private IRequestHandler getRequestHandler(
               request.getGroupByList());
       Optional<String> source =
           ExpressionContext.getSingleSourceForAllAttributes(expressionContext);
-      if (source.isPresent() && EDS.toString().equals(source.get())) {
+      if ((source.isPresent() && EDS.toString().equals(source.get())) || !hasTimeRange(request)) {
```
skjindal93, line 134: nit: can we add a small unit test?

PR #134, opened 2022-06-06
Added boolen flag to disable time interval alignment in explore query

singhalprerana:
Explore query, by default, aligns the time range to the queried time interval.
However, in certain cases, results based on the queried time-range are required, which means no alignment.
Added a field in explore-request to allow this.

### DIFF
--- a/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
+++ b/gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
@@ -32,6 +32,7 @@ message ExploreRequest {
   int32 offset = 21;
 
   string space_id = 22;
+  bool disable_time_interval_alignment = 23; // false by default for backward compatibility
 }
 
 message ExploreResponse {
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsRequestHandler.java
@@ -50,7 +50,9 @@ public class TimeAggregationsRequestHandler extends RequestHandler {
     QueryRequest.Builder builder = QueryRequest.newBuilder();
 
     // 1. Align the startTime and endTime with period boundaries if there are TimeAggregations
-    request = createPeriodBoundaryAlignedExploreRequest(request);
+    if (!request.getDisableTimeIntervalAlignment()) {
+      request = createPeriodBoundaryAlignedExploreRequest(request);
+    }
 
     // 2. Add filter
     builder.setFilter(


gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
```diff
@@ -32,6 +32,7 @@ message ExploreRequest {
   int32 offset = 21;
 
   string space_id = 22;
+  bool disable_time_interval_alignment = 23; // false by default for backward compatibility
```
aaron-steinfeld, line 35: A boolean flag controlling behavior is a big code smell - can you elaborate on why we would want inconsistent behavior?

singhalprerana: we want the given time-range to be equally divided into intervals and get the data for such intervals instead of aligning the intervals to timestamp.. @arjunlalb could u please elaborate?

aaron-steinfeld: So two threads of this conversation.

First, regardless of whether we align or not, we want the UI to be consistent. If we change the behavior to not align, the entire UI should behave that way, rather than different widgets picking and choosing. That's my main concern with this PR, that we don't want to expose this to the API but rather leave it an implementation detail and if we choose to change it, change it for all callers.

Now, as to why alignment was added in the first place. I don't recall the specific reason (I can try to dig into history), but I can think of a bunch of UX issues that removing this would introduce:
1. Users think in wall clock time. If you look at metrics over an hour interval, you want to see it from 6-7PM rather than from 635-735PM. When drawing a graph this is less of a concern, but when inspecting it, more so. As I write this though, I'm curious how this works for time zones like IST that are not hour-aligned with UTC. I would guess a 630-730PM interval? This one is a bit wishy-washy, but if I had to bet money - this is why it was done.
2. Data consistency. If we move the time range and that recalculates what the intervals are, the actual plotted data will change on refresh. Historical data will be re-bucketed depending on the time range leading to potentially different graph shapes for the same data after refreshing. IMO, this is actually the biggest issue with removing it.
3. Data liveness. Our data is close to real-time, but not exactly. There's always going to be some delay, usually on the order of a few minutes. If at 735PM I query and create a 535-635 and 635-735 bucket, the latter bucket will have less data in it which will lead to an artificial drop off for summation and count-type queries. We want every bucket to be full, and using an exact alignment will prevent that. Again, thinking through this I would guess we'd hit this issue today too in more limited circumstances, such as a query made right at the boundary (e.g. I query at 7pm and get a 6-7pm bucket).

So I'm not saying we can't reconsider, but what I am saying is there's a lot of UX thinking involved with how this is done, and we should have a more in depth discussion around this rather than just changing it because it feels wrong.

arjunlalb: Thanks for the explanation Aaron. 

I agree on the point that we should consistently apply this behaviour everywhere. Do you think adding this toggle and querying for it based on an FF in UI will work in the interim? That'll allow stakeholders to evaluate the UX based on the points you elaborated.

My thoughts on the reasons you mentioned :

1. Yes, users tend to think in wall clock time, but I don't think that should determine this behaviour. If we use that logic, we should not be using referential time ranges like `last x minutes/hours`. We should instead allow user to choose only full wall clock like time intervals. So, we are saying everywhere we show data, we'll respect the last x minutes/hours promise, but not when we show interval data? That is inconsistent.
2. I think of this is as an extension of the previous point. The data and the time range will change when you hit refresh. Isn't that the point of hitting refresh? get the latest data? We cannot maintain context if the user hits refresh after the endTime has passed anyway, the entire data shown is new. I would argue we should just respect the user's choice to fetch the latest data and give them that.
3. To solve this, I think its better if we decide that we'll always return `timeRange/interval - 1` buckets avoiding the last piece that cannot be completely aggregated yet.

We should have Product and UX look into this.

aaron-steinfeld: >  Do you think adding this toggle and querying for it based on an FF in UI will work in the interim? 

Would rather do it as helm config in gateway.

>  So, we are saying everywhere we show data, we'll respect the last x minutes/hours promise, but not when we show interval data? That is inconsistent.

I'm not sure I agree, but certainly worth a discussion.

> Isn't that the point of hitting refresh? get the latest data? 

Here however, I definitely disagree. Refresh should be moving the graph to the right (that is, forward in time) it shouldn't be rebucketing and potentially changing its shape, moving peaks etc.

> I think its better if we decide that we'll always return timeRange/interval - 1 buckets avoiding

This is weird too though - so in the 635-835 case, we'd only return one interval, 635-735? If we're only going to get one interval, I'd rather get 7-8 in that case. Neither is good here.

aaron-steinfeld: > We should have Product and UX look into this.

Yeah, could you set up some conversations if you think this is a priority?

singhalprerana: Aren't these 2 different use-cases, rather requirements.? 
First, the existing one, where we want to show the charts for intervals which should be fixed on refresh - more of a left moving kind of chart where buckets are fixed and new ones are added with every passing interval. 
Second, the reason behind this change, is given a group of items in a given time range, divide it into equal time divisions to show the distribution of the items across the selected time range. There is no question of refresh here or showing fixed buckets wrt universal time. It's merely a requirement of dividing calls based on time. For example, if I saw 50k security threats on Monday between 5.35am and 7.35pm in my timezone when I know my app saw peak traffic, I want to see the distribution of the threats across this time range.. I just want to divide this into 14 hours and see the distribution.. later if I want even further distribution into some hour by minute, I can do that.. What I mean is this is a division requirement where the exact timestamp bucketing is irrelevant.
Both these requirements are valid and unrelated. What we use in our product today is generally charts where fixed bucket segments make sense. But this new requirement of request division is not a chart visualization - it's more of a classification of requests to allow further drilling down into the data.
@aaron-steinfeld Not sure if I was able to explain my thoughts, but we can have a sync-up to finalize on this as this is a blocking requirement for the new Threat Activity dashboard. ref UI where this needs to be used - https://app-dev.traceable.ai/threat-activity/activity?filter=name_eq_MSSQL%20code%20execution%20and%20information%20gathering%20attempts&filter=type_eq_Signature:%20SQL%20Injection&filter=securityScore_eq_4&filter=threatCategory_eq_Malicious%20Activities&filter=category_eq_SECURITY&time=1w (Request Timeline part)

singhalprerana:
Not required anymore

PR #146, opened 2022-09-19
chore: rename DISTINCT function type to DISTINCT_ARRAY

SrikarMannepalli:
## Description
This PR renames the DISTINCT function type to DISTINCT_ARRAY. Making a breaking proto change as it is not being consumed anywhere yet.

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
@@ -416,16 +416,14 @@ public class QueryAndGatewayDtoConverter {
           columns.forEach(e -> builder.addArguments(convertToQueryExpression(e)));
           break;
         }
+      case DISTINCT_ARRAY:
+        throw new IllegalArgumentException(
+            format(
+                "Aggregation by the function type: %s is not supported by query service currently",
+                function.getFunction().name()));
       default:
         {
-          String functionName = function.getFunction().name();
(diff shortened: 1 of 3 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -416,7 +418,14 @@ private static Function convertToQueryFunction(
         }
       default:
         {
-          builder.setFunctionName(function.getFunction().name()).setAlias(function.getAlias());
+          String functionName = function.getFunction().name();
+          if (DISTINCT_ARRAY.name().equals(functionName)) {
```
skjindal93, line 422: Can we just add it as a switch case above and throw the exception?

SrikarMannepalli: Right. Updated it now.

surajpuvvada:
@kotharironak does this impact postgressql layer ?

aaron-steinfeld:
> @kotharironak does this impact postgressql layer ?

Already had it, there's just been some renaming going on. Both are supported and map to the same thing.

ES Support was added https://github.com/hypertrace/entity-service/commit/d478297824bc7a7b4b1688a1dce387e13e867040

Doc store/postgres support renamed DISTINCT to DISTINCT_ARRAY in
https://github.com/hypertrace/document-store/pull/128
then fixed compatibility by restoring DISTINCT in https://github.com/hypertrace/document-store/pull/129

PR #152, opened 2022-11-08
chore | introducing edge filters in get entities interaction

aman-bansal:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +275,82 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+
+    Map<String, QueryRequest> queryRequests = new HashMap<>();
+
+    // In future we could send these queries in parallel to QueryService so that we can reduce the
+    // response time.
+    for (String e : entityTypes) {
+      DomainEntityType otherEntityType = DomainEntityType.valueOf(e.toUpperCase());
+
+      // Get the filters from the interactions request to 'AND' them with the timestamp filter.
+      Filter.Builder filterCopy = Filter.newBuilder(protoTypeFilter);
```
skjindal93, line 331: We can avoid building the filter on L321, and directly add the child filters to it, instead of creating a copy

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +275,82 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+      filterCopy.addChildFilter(
+          convertToQueryFilter(interactionsRequest.getFilter(), otherEntityType));
+
+      QueryRequest.Builder builderCopy = QueryRequest.newBuilder(protoType);
```
skjindal93, line 335: Similarly, we can avoid building the query request on L320, and directly set filter

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +275,82 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+      builderCopy.setLimit(DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+      queryRequests.put(e, builderCopy.build());
+    }
+
+    return queryRequests.entrySet().stream()
+        .map(
+            e ->
+                new EntityInteractionQueryRequest(
+                    incoming, e.getKey(), interactionsRequest, e.getValue()))
+        .collect(Collectors.toList());
```
skjindal93, line 351:
```suggestion
        .collect(Collectors.toUnmodifiableList());
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,90 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+                return;
+              }
+              resultSetChunk
+                  .getRowList()
+                  .forEach(
+                      row -> {
+                        List<String> idColumns =
+                            getEntityIdColumnsFromInteraction(
+                                DomainEntityType.valueOf(entitiesRequest.getEntityType()),
+                                !qsResponse.getRequest().isIncoming());
```
skjindal93, line 196: A bit confused. Why are we building entity ids only for outgoing edges?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,90 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+                        EntityKey key =
+                            EntityKey.of(
+                                IntStream.range(0, idColumns.size())
+                                    .mapToObj(value -> row.getColumn(value).getString())
+                                    .toArray(String[]::new));
+                        entityKeys.add(key);
+                      });
+            });
+    return entityKeys;
```
skjindal93, line 205:
```suggestion
    return Collections.unmodifableList(entityKeys);
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -149,6 +156,8 @@ public EntitiesResponse getEntities(
     EntitiesRequest preProcessedRequest =
         requestPreProcessor.process(originalRequest, entitiesRequestContext);
 
+    preProcessedRequest = addEntityIdFromInteractionFilter(requestContext, preProcessedRequest);
```
skjindal93, line 159:
```suggestion
    preProcessedRequest = addEntityIdsFromInteractionFilter(requestContext, preProcessedRequest);
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,96 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+            .collect(Collectors.toUnmodifiableList());
+
+    Set<EntityKey> entityKeys = new HashSet<>();
+    queryRequestCompletableFutures.forEach(
+        queryRequestCompletableFuture -> {
+          EntityInteractionQueryResponse qsResponse = queryRequestCompletableFuture.join();
+          entityKeys.addAll(parseInteractionIdsResponse(entitiesRequest, qsResponse));
+        });
+
+    return entityKeys;
```
skjindal93, line 182:
```suggestion
    return Collections.unmodifiableSet(entityKeys);
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -235,6 +246,61 @@ public BulkUpdateEntitiesResponse bulkUpdateEntities(
...
+      entities = entities.subList(0, DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+    }
+    List<String> entityIds =
+        entities.stream().map(EntityKey::toString).collect(Collectors.toUnmodifiableList());
+
+    List<String> idAttributeIds =
+        AttributeMetadataUtil.getIdAttributeIds(
+            metadataProvider, entityIdColumnsConfigs, requestContext, entityType);
+
+    List<Filter> entityIdsInFilter =
```
skjindal93, line 287: A bit confused. If we have multiple id attributes for an entity, why are we adding the same id attribute values for all the multiple id attributes

skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -235,6 +246,61 @@ public BulkUpdateEntitiesResponse bulkUpdateEntities(
...
+      RequestContext requestContext, EntitiesRequest preProcessedRequest) {
+    Set<EntityKey> entityKeys =
+        this.interactionsFetcher.fetchInteractionsIdsIfNecessary(
+            requestContext, preProcessedRequest);
+    if (!entityKeys.isEmpty()) {
+      Filter entityIdsInFilter =
+          createEntityIdInFilter(requestContext, preProcessedRequest.getEntityType(), entityKeys);
+
+      EntitiesRequest.Builder preProcessRequestBuilder = preProcessedRequest.toBuilder();
+      if (preProcessRequestBuilder.getFilter().equals(Filter.getDefaultInstance())) {
```
aaron-steinfeld, line 259: nit: prefer hasFilter rather than comparing to default instance

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -235,6 +246,61 @@ public BulkUpdateEntitiesResponse bulkUpdateEntities(
...
+                .build());
+      }
+      preProcessedRequest = preProcessRequestBuilder.build();
+    }
+    return preProcessedRequest;
+  }
+
+  private Filter createEntityIdInFilter(
+      RequestContext requestContext, String entityType, Set<EntityKey> entityKeys) {
+    List<EntityKey> entities = new ArrayList<>(entityKeys);
```
aaron-steinfeld, line 276: These aren't entities, they're keys - please update naming.

Also confused on the next few lines - if we received more keys than we can return entities, why would we drop them? This isn't the only filter - other filters can still bring the result beneath the limit (and even if it was, that's what the limit is for - we shouldn't be enforcing in code)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
aaron-steinfeld: Yeah this looks like a bug to me. If anything is using multiple ID attributes (not sure if that's the case today), this would fail because we're assuming each one should contain the same value, while the intention was that each part contains a portion of the ID.

aaron-steinfeld (dismissed):
not done reviewing, but I don't think this is quite ready to merge

aman-bansal:
Regarding this https://github.com/hypertrace/gateway-service/pull/152#discussion_r1020382785 I assumed the id filter evaluation based on here org.hypertrace.gateway.service.common.datafetcher.EntityInteractionsFetcher#createFilterForEntityKeys. Ideally this should never happen. No one is using multi id attributes but i believe this is legacy code written in that way

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
aman-bansal: Okay consider this case, lets say there are 5 interaction requests. Now each query will fetch 10k entity Ids. So in the worst case, the entity id filter will have 5 x 10k entity ids as the filter. I put this check to safeguard the 10k limit over the filter

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+            AttributeMetadataUtil.getTimestampAttributeId(metadataProvider, requestContext, SCOPE),
+            request.getStartTimeMillis(),
+            request.getEndTimeMillis()));
+
+    this.buildSpaceQueryFilterIfNeeded(requestContext, request.getSpaceId())
+        .ifPresent(childFilters::add);
+
+    List<String> idColumns =
+        getEntityIdColumnsFromInteraction(
+            DomainEntityType.valueOf(request.getEntityType()), !incoming);
```
aaron-steinfeld, line 306: Looks like @skjindal93 had the same questions here I do - sounds like it deserves a comment or better yet, improve the code. For example, rather than using a boolean called incoming, this method could be updated to take an enum indicating whether we want to get the caller or callee ID columns.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+      EntitiesRequest request,
+      InteractionsRequest interactionsRequest,
+      boolean incoming) {
+    Set<String> otherEntityTypes = getOtherEntityTypes(interactionsRequest.getFilter());
+    if (otherEntityTypes.isEmpty()) {
+      return Collections.emptyList();
+    }
+
+    List<Filter> childFilters = new ArrayList<>();
+    childFilters.add(
```
aaron-steinfeld, line 295: nit: the more you can avoid mutating, the more readable the code becomes

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+        getEntityIdColumnsFromInteraction(
+            DomainEntityType.valueOf(request.getEntityType()), !incoming);
+
+    // Group by the entity id column first, then the other end entity type for the interaction.
+    List<org.hypertrace.core.query.service.api.Expression> idExpressions =
+        idColumns.stream()
+            .map(QueryRequestUtil::createAttributeExpression)
+            .collect(Collectors.toUnmodifiableList());
+
+    // adding empty selections
```
aaron-steinfeld, line 314: The selection isn't empty - it's a count. Why are we adding it? Initially I thought it was for passing validation, but I see we add more later. Comments are typically more helpful if they explain "why" rather than "what".

Also if leaving it, there's no reason to use a mutable list here since the following line is the only time we add to it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+    List<org.hypertrace.core.query.service.api.Expression> selections = new ArrayList<>();
+    selections.add(
+        QueryRequestUtil.createCountByColumnSelection(
+            Optional.ofNullable(idColumns.get(0)).orElseThrow()));
+
+    Map<String, QueryRequest> queryRequests = new HashMap<>();
```
aaron-steinfeld, line 320: If I'm reading how the map is used, we're doing a for loop, putting in the map then transforming the map at the end. Instead, stream the entity types and collect requests.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+    for (String e : otherEntityTypes) {
+      DomainEntityType otherEntityType = DomainEntityType.valueOf(e.toUpperCase());
```
aaron-steinfeld, line 322: Please try to remove any new code (and if you can/need to, update any existing code) using this enum. It'll break for new entity types.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -191,6 +281,72 @@ private EntityInteractionQueryResponse executeQueryRequest(
...
+
+      // Get the filters from the interactions request to 'AND' them with the timestamp
+      // defaultFilter.
+      Filter.Builder filterBuilder = Filter.newBuilder().addAllChildFilter(childFilters);
+      filterBuilder.addChildFilter(
```
aaron-steinfeld, line 327: nit: chain rather than imperative
```suggestion
      Filter.Builder filterBuilder = Filter.newBuilder()
          .addAllChildFilter(childFilters)
          .addChildFilter(
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,96 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+  private boolean hasInteractionFilters(org.hypertrace.gateway.service.v1.common.Filter filter) {
+    if (ExpressionReader.isSimpleAttributeSelection(filter.getLhs())) {
+      String attributeId =
+          ExpressionReader.getAttributeIdFromAttributeSelection(filter.getLhs()).orElseThrow();
+
+      return !StringUtils.equals(attributeId, FROM_ENTITY_TYPE_ATTRIBUTE_ID)
+          && !StringUtils.equals(attributeId, TO_ENTITY_TYPE_ATTRIBUTE_ID);
+    }
+
+    boolean hasFilters = false;
```
aaron-steinfeld, line 133: nit - avoid the variable and if check. We're adding extra lines of code and things to hold in our mind for no reason - pretty sure this is equivalent to these 6 lines:
```java
return filter.getChildFilterList().stream().anyMatch(this::hasInteractionFilters);
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
aaron-steinfeld: Yes, that's a problem - but it's a problem with the general approach that I had mentioned earlier and this isn't a good safeguard for it, since it just breaks the filter logic.

I could just as easily have an edge filter that's super generic that matches everything or almost everything - effectively the edges shouldn't be filtered.

So imagine if I had a querty that said "I want a service with ID = "xyz" and outgoing api calls where the api name matches ".*". By pre-processing the outgoing api calls to get a list of service IDs that match first, and appending that to the filter, if XYZ isn't in the first 10K we'd be forming a filter that returns no results. Hope that makes sense.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
aaron-steinfeld: > I assumed the id filter evaluation based on here org.hypertrace.gateway.service.common.datafetcher.EntityInteractionsFetcher#createFilterForEntityKeys. Ideally this should never happen. No one is using multi id attributes but i believe this is legacy code written in that way

Threading the result here since lots of comments.

If this should never happen, then we should ideally fix gateway service to remove support - or fix the code to support it, although not sure how that would work. If we can't do either, at a minimum make the code loudly error out if it ever runs into this scenario. Right now, we're making the code support it, but the logic doesn't - so if it were ever used it would lead to very annoying to track down bugs.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
aman-bansal: I have removed this trim method

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -271,34 +269,27 @@ private EntitiesRequest addEntityIdsFromInteractionFilter(
...
-                        .setOperator(IN)
-                        .setRhs(QueryExpressionUtil.getLiteralExpression(entityIds).build())
-                        .build())
-            .collect(Collectors.toUnmodifiableList());
-    if (entityIdsInFilter.size() == 1) {
-      return entityIdsInFilter.get(0);
-    } else {
-      return Filter.newBuilder().setOperator(OR).addAllChildFilter(entityIdsInFilter).build();
+    if (idAttributeIds.size() != 1) {
+      throw new RuntimeException("Entity Type " + entityType + " should have single ID Attribute");
```
aaron-steinfeld, line 285: probably a Status.UNIMPLEMENTED would be appropriate here

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -666,8 +657,7 @@ private void parseResultSet(
                     .toArray(String[]::new));
 
         List<String> otherIdColumns =
-            getEntityIdColumnsFromInteraction(
-                DomainEntityType.valueOf(otherEntityType.toUpperCase()), incoming);
+            getEntityIdColumnsFromInteraction(otherEntityType.toUpperCase(), incoming);
```
aaron-steinfeld, line 660: I'm guessing from this toUpperCase we're just hiding a conversion to enum - so the problem isn't solved. If fixing that is out of scope, then we'd have to introduce an alternative. If it's not prohibitive, maybe it makes sense to create a separate PR to be merged first that fixes this.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,96 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+        .forEachRemaining(
+            resultSetChunk -> {
+              if (resultSetChunk.getRowCount() < 1) {
+                return;
+              }
+              resultSetChunk
+                  .getRowList()
+                  .forEach(
+                      row -> {
+                        List<String> idColumns =
```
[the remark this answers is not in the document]
aaron-steinfeld: I'm not sure I'm following - all I was referring to here is that this function we're evaluating each time doesn't take either loop var as an input, so can be evaluated once outside the double nested loop.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
aaron-steinfeld: sure let's verify. That would be **very** surprising if it required an aggregation in a group by request.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -121,6 +121,96 @@ private List<String> getEntityIdColumnsFromInteraction(
...
+                            getEntityIdColumnsFromInteraction(
+                                DomainEntityType.valueOf(entitiesRequest.getEntityType()),
+                                !qsResponse.getRequest().isIncoming());
+                        EntityKey key =
+                            EntityKey.of(
+                                IntStream.range(0, idColumns.size())
```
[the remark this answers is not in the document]
aaron-steinfeld: The whole function doesn't have to be shared, but large parts likely can be - I'm sure there's some useful refactoring you can do here.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
aman-bansal: I am changing this to entityType only. No uppercase, I feel like that was only a precautionary check before also.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/entity/EntityServiceInteractionRequestTest.java
```diff
@@ -411,6 +413,18 @@ private void mockQueryServiceRequestForServiceCountWithInteractionFilters(
                 .iterator());
   }
 
+  public org.hypertrace.core.query.service.api.Expression createStringSetLiteralExpression(
```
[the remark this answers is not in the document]
aman-bansal: Ignore this for a while. Pr is in unstable state. Tests are failing i am trying to understand the correct way to fix the issues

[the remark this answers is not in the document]
aaron-steinfeld: Sure - useful to convert it to draft in that case to indicate to reviewers that it's not ready for review. Haven't looked at the test failures, but based on some of what I'm seeing in this file if there are ordering concerns, it's likely due to the use of unordered data structures in the implementation. Unless there's a specific reason to use a set, I would suggest using ordered structures everywhere - makes things repeatable and more testable.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
aman-bansal: I was wrong this isn't needed. removing this

skjindal93 (dismissed):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/datafetcher/EntityInteractionsFetcher.java
```diff
@@ -164,16 +164,18 @@ public Set<EntityKey> fetchInteractionsIdsIfNecessary(
...
-        .map(CompletableFuture::join)
-        .map(response -> parseInteractionIdsResponse(entitiesRequest, response))
-        .flatMap(Collection::stream)
-        .collect(Collectors.toUnmodifiableSet());
+    LinkedHashSet<EntityKey> entityKeys =
+        queryRequestCompletableFutures.stream()
+            .map(CompletableFuture::join)
+            .map(response -> parseInteractionIdsResponse(entitiesRequest, response))
+            .flatMap(Collection::stream)
+            .collect(Collectors.toCollection(LinkedHashSet::new));
```
aaron-steinfeld, line 172: nit - are you trying to use the set just to make sure there aren't dupes? If your goal is for it to be both ordered and no dupes, it would be clearer to just do:
```
.distinct()
.collect(Collectors.toUnmodifiableList());
```

aaron-steinfeld (dismissed):

skjindal93 (approved):

PR #154, opened 2022-11-18
fix | fixing the pre process filter request

aman-bansal:


skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -253,7 +253,10 @@ private EntitiesRequest addEntityIdsFromInteractionFilter(
           createEntityKeysInFilter(requestContext, preProcessedRequest.getEntityType(), entityKeys);
 
       EntitiesRequest.Builder preProcessRequestBuilder = preProcessedRequest.toBuilder();
-      if (preProcessRequestBuilder.hasFilter()) {
+      // This check is required and we cannot convert this to hasFilter check.
```
aaron-steinfeld, line 256: even with this change which will be more robust - we should fix the upstream bug too. shouldn't be hard to find

PR #156, opened 2022-12-05
fix(explore): time aggregations group by

skjindal93:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -36,6 +36,7 @@
 
 public class RequestHandler implements RequestHandlerWithSorting {
   private static final Logger LOG = LoggerFactory.getLogger(RequestHandler.class);
+
```
sarthak77, line 39: Is this intentional?

skjindal93: yes, I prefer to keep a gap between static and non static class variables

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandlerTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.hypertrace.gateway.service.v1.explore.ExploreRequest;
+import org.hypertrace.gateway.service.v1.explore.ExploreResponse;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+public class TimeAggregationsWithGroupByRequestHandlerTest {
+  @Mock private AttributeMetadataProvider attributeMetadataProvider;
+  @Mock private RequestHandler normalRequestHandler;
+  @Mock private TimeAggregationsRequestHandler timeAggregationsRequestHandler;
```
sarthak77, line 38: nit : can remove mock annotation.

sarthak77 (dismissed):
LGTM

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandlerTest.java
suresh-prakash: +1

suresh-prakash (dismissed):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -1,31 +1,54 @@
...
 
   TimeAggregationsWithGroupByRequestHandler(
       QueryServiceClient queryServiceClient, AttributeMetadataProvider attributeMetadataProvider) {
+    this.attributeMetadataProvider = attributeMetadataProvider;
     this.normalRequestHandler = new RequestHandler(queryServiceClient, attributeMetadataProvider);
     this.timeAggregationsRequestHandler =
         new TimeAggregationsRequestHandler(queryServiceClient, attributeMetadataProvider);
   }
 
+  @VisibleForTesting
```
aaron-steinfeld, line 42: 😢 (this version is actually better, but I'd rather remove the old constructor and make the actual code use this one too)

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandlerTest.java
aaron-steinfeld: opposite nit: remove the setup step and leave the mock annotations (you're missing the class annotation which is why the mocks aren't working)

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -102,42 +129,143 @@ private ExploreRequest buildTimeAggregationsRequest(
...
+
+    if (!resultKeyToAttributeMetadataMap.containsKey(groupByResultName)) {
+      return Optional.empty();
+    }
+
+    AttributeMetadata groupByAttributeMetadata =
+        resultKeyToAttributeMetadataMap.get(groupByResultName);
+    // RHS value of in clause filter should always be an array to apply IN filter clause on groupBy
+    // expression
+    Value.Builder valueBuilder = Value.newBuilder();
```
aaron-steinfeld, line 173: can we move this out from being inline in a giant method. Ideally into its own class.

aaron-steinfeld: I realize now that this is really the whole point of this method, but the generally point I think remains - it's a ~100 line method with multiple giant switches in it. Must be a way to clean that up.

gateway-service-impl/src/test/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandlerTest.java
sarthak77: >remove the setup step

I think we need to open and close the mocks then.

aaron-steinfeld: Mockito does that automatically if you use 
```
@ExtendWith(MockitoExtension.class)
```
It also makes tests strict and all kinds of other goodies - we really shouldn't be writing new tests without it.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -102,42 +128,148 @@ private ExploreRequest buildTimeAggregationsRequest(
...
+        remapAttributeMetadataByResultName(
+            originalRequestContext.getExploreRequest(), attributeMetadataMap);
+
+    if (!resultKeyToAttributeMetadataMap.containsKey(groupByResultName)) {
+      return Optional.empty();
+    }
+
+    AttributeMetadata groupByAttributeMetadata =
+        resultKeyToAttributeMetadataMap.get(groupByResultName);
+    // RHS value of in clause filter should always be an array to apply IN filter clause on groupBy
```
sarthak77, line 170: This comment is present twice. Ig can be removed from here.

aaron-steinfeld (approved):

PR #158, opened 2022-12-06
API implementation for bulk updating entities by filter

suresh-prakash:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
```diff
@@ -289,6 +304,43 @@ public BulkUpdateEntitiesResponse bulkUpdateEntities(
...
+    final Converter<BulkUpdateAllMatchingFilterResponse, BulkUpdateAllMatchingEntitiesResponse>
+        responseConverter =
+            injector.getInstance(
+                Key.get(
+                    new TypeLiteral<
+                        Converter<
+                            BulkUpdateAllMatchingFilterResponse,
+                            BulkUpdateAllMatchingEntitiesResponse>>() {}));
+
+    final BulkUpdateAllMatchingFilterResponse response =
```
skjindal93, line 337: Should we have a try/catch here?

skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/EntityService.java
suresh-prakash: It is there in the caller of this method.

PR #159, opened 2022-12-09
Reduce the attribute cache interval for custom attributes to get reflected faster

suresh-prakash:


skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/converter/UpdateOperationConverter.java
```diff
@@ -21,10 +24,15 @@
...
-        .setAttribute(columnIdentifierConverter.convert(source.getAttribute()))
-        .setOperator(operatorConverter.convert(source.getOperator()))
-        .setValue(literalConstantConverter.convert(source.getValue()))
-        .build();
+    final Builder builder =
+        AttributeUpdateOperation.newBuilder()
+            .setAttribute(columnIdentifierConverter.convert(source.getAttribute()))
+            .setOperator(operatorConverter.convert(source.getOperator()));
+
+    if (source.getOperator() != OPERATOR_UNSET) {
```
aaron-steinfeld, line 32: If the operator is unset, why are we setting it on line 30?

aaron-steinfeld: But why do we need to convert and set the operator? Is unset a special value? Usually if we get an unset value, we don't want to blindly set it in the proto, we want to leave it unset to support presence detection.

suresh-prakash: I think, I understood why we are confused here. 
This is a deliberately created enum value (not the proto generated one)
image

aaron-steinfeld: Ah - got it, yes I thought this was the unspecified case! thanks for following up.

PR #167, opened 2023-07-12
chore | adding support for non live entities in explore

aman-bansal:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityRequestHandler.java
```diff
@@ -90,15 +94,18 @@ public ExploreResponse.Builder handleRequest(
       requestContext.setHasGroupBy(true);
     }
 
-    Set<String> entityIds = getEntityIds(requestContext, exploreRequest);
     ExploreResponse.Builder builder = ExploreResponse.newBuilder();
-
-    if (entityIds.isEmpty()) {
-      return builder;
+    Set<String> entityIds = new HashSet<>();
+    if (!exploreRequest.getIncludeNonLiveEntities()) {
```
skjindal93, line 99: A bit confused. Where are we checking for fetching results time agnostically?

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
```diff
@@ -32,6 +32,7 @@ message ExploreRequest {
   int32 offset = 21;
 
   string space_id = 22;
+  bool include_non_live_entities = 23;
```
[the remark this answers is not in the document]
kotharironak: Additional information for a given context, should be align as `oneOf` operator?
For now, it is`EntityOptions`. Later if we need to add for span or API_Trace, it will be `ApiTraceOptions`.

Also this is PR changes the API signature, so, it should be mark as `feat` or something else. Not as `chore`.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -124,16 +121,19 @@ private IRequestHandler getRequestHandler(
               request.getTimeAggregationList(),
               request.getOrderByList(),
               request.getGroupByList());
+
+      boolean allFieldsOnEDS =
```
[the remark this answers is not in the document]
aman-bansal: Okay so here is my confusion, if there is request.getIncludeNonLiveEntities(), then all are needed to be on EDS not as the only option. Say all fields are on QS, EDS, then by default QS would be selected because it will skip all the checks.  

Thats why needed these two checks and that's why field is named allFieldsOnEDS

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityRequestHandler.java
aman-bansal: because say there are only non-live entities. then if we don't put this check results won't be returned. Secondly, if non live entities are selected filter and group by wont work properly because we are putting entity ids filter which are in live traffic only

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
```diff
@@ -32,6 +32,17 @@ message ExploreRequest {
...
+}
+
+message ContextOption {
+  oneof option {
+    EntityOption entity_option = 1;
+  }
+}
+
+message EntityOption {
+  bool include_non_live_entities = 1;
```
skjindal93, line 45: We should ideally get rid of the `optional` flags for start time and end time

aman-bansal: let me check if those are breaking change or not

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -124,16 +121,19 @@ private IRequestHandler getRequestHandler(
...
+          ExpressionContext.areAllFieldsOnlyOnCurrentDataSource(expressionContext, EDS.name());
+      if (allFieldsOnEDS && request.getIncludeNonLiveEntities()) {
+        return entityRequestHandler;
+      }
+
       Optional<String> source =
           ExpressionContext.getSingleSourceForAllAttributes(expressionContext);
```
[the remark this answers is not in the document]
aman-bansal: I dont think we can skip this altogether. I have written all the cases in comment and this falls under Case 3.

skjindal93 (approved):

PR #168, opened 2023-07-12
feat: enhance the spans, traces api for optimisation

kotharironak:
## Description
Currently, the response of both the spans and traces APIs includes the total field by default. Here is an example of the response message: 

e.g
```
message TracesResponse {
  repeated Trace traces = 1;
  int32 total = 2;
}
```

As a result, the API executes two database queries:
- One to fetch the requested data.
- Another to calculate the total.

In order to support optimization where the total value is not required to be shown in the UI, we need to provide a way for clients to request not to calculate the total.

In this pull request (PR), we are enhancing the request protobuf definitions for both the spans and traces APIs by adding an additional request parameter called `fetch_total`. This parameter can be used by clients to indicate whether or not the total value should be calculated and included in the API response.

Refer API change :  https://github.com/hypertrace/gateway-service/pull/169

### Testing
Added unit tests

Note: Graphql Changes that will request based on UI request to fetch_total will be added as separate PR.

### Checklist:

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
@@ -71,12 +71,16 @@ public class SpanService {
       Map<String, AttributeMetadata> attributeMap =
           attributeMetadataProvider.getAttributesMetadata(context, AttributeScope.EVENT.name());
       SpansResponse.Builder spanResponseBuilder = SpansResponse.newBuilder();
-      CompletableFuture<Collection<SpanEvent>> filteredSpanEventsFuture =
-          CompletableFuture.supplyAsync(
-              () -> filterSpans(context, request, attributeMap), queryExecutor);
-
-      spanResponseBuilder.setTotal(getTotalFilteredSpans(context, request));
-      spanResponseBuilder.addAllSpans(filteredSpanEventsFuture.join());
+      if (request.getFetchTotal()) {
+        CompletableFuture<Collection<SpanEvent>> filteredSpanEventsFuture =
+            CompletableFuture.supplyAsync(
+                () -> filterSpans(context, request, attributeMap), queryExecutor);
+        spanResponseBuilder.setTotal(getTotalFilteredSpans(context, request));
+        spanResponseBuilder.addAllSpans(filteredSpanEventsFuture.join());
+      } else {
+        // if total is not request, we can fetch the spans in the same thread.
+        spanResponseBuilder.addAllSpans(filterSpans(context, request, attributeMap));
+      }
 
       SpansResponse response = spanResponseBuilder.build();
       LOG.debug("Span Service Response: {}", response);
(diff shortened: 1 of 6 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
```diff
@@ -185,46 +185,50 @@ private QueryRequest.Builder createQueryWithFilter(
   }
 
   private int getTotalFilteredSpans(RequestContext context, SpansRequest request) {
```
satish-mittal, line 187: This could return Optional<Integer> and let the caller decide whether to return -1 or something else.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
```diff
@@ -185,46 +185,50 @@ private QueryRequest.Builder createQueryWithFilter(
...
-      if (chunk.getRowCount() != 1 && chunk.getResultSetMetadata().getColumnMetadataCount() != 1) {
-        LOG.error(
-            "Count the Api Traces total returned in multiple row / column. "
-                + "Total Row: {}, Total Column: {}",
-            chunk.getRowCount(),
-            chunk.getResultSetMetadata().getColumnMetadataCount());
-        break;
-      }
+    int total = -1;
+    if (request.getFetchTotal()) {
```
skjindal93, line 189: We should add the check in the caller rather than here. We can either completely avoid setting total (which will make the total by default as 0), if the `request.getFetchTotal()` is false, or we can set it to -1

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/trace/TracesService.java
```diff
@@ -172,51 +172,55 @@ List<Trace> filterTraces(
...
-            .build();
-
-    Iterator<ResultSetChunk> resultSetChunkIterator =
-        queryServiceClient.executeQuery(context, queryRequest);
-    while (resultSetChunkIterator.hasNext()) {
-      ResultSetChunk chunk = resultSetChunkIterator.next();
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Received chunk: " + chunk.toString());
+    int total = -1;
+    if (request.getFetchTotal()) {
```
skjindal93, line 176: Can we move the check to the caller instead calling the `getTotalFilteredTraces`?

kotharironak: To keep the api compatible, 
- we will first merge API change as follow
- change the consumer
- provide new api implementation

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/trace/TracesService.java
```diff
@@ -172,51 +172,55 @@ List<Trace> filterTraces(
...
-            .setLimit(1)
-            .build();
-
-    Iterator<ResultSetChunk> resultSetChunkIterator =
-        queryServiceClient.executeQuery(context, queryRequest);
-    while (resultSetChunkIterator.hasNext()) {
-      ResultSetChunk chunk = resultSetChunkIterator.next();
-      if (LOG.isDebugEnabled()) {
-        LOG.debug("Received chunk: " + chunk.toString());
+    int total = -1;
```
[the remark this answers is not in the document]
kotharironak: Not needed, see new API def : https://github.com/hypertrace/gateway-service/pull/169

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
kotharironak: Not needed with new api def.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/span/SpanService.java
kotharironak: Done.

satish-mittal (approved):

skjindal93 (approved):

PR #170, opened 2023-07-12
ENG-33049 - Make total computation optional.

surajpuvvada:
Note that if total is not computed the response will have a default value of 0. There is no way of knowing if it was unset as opposed to it is really 0


kotharironak:
Duplicate of https://github.com/hypertrace/gateway-service/pull/169. Closing this one.

PR #171, opened 2023-07-25
chore:  Handle scenario when query has a group by

saxenakshitiz:
Changes in query going to Pinot -
1. Now limit is passed down as is for a order by query
2. If group limit is also provided, then min of group-limit and limit is used
3. If query has `includeRest: true`, then rest group is fetched if and only if there is limit available to fetch more results.
4. We don't support order by query with offset, and such query will results in UnsupportedOperationException


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -98,17 +98,20 @@ QueryRequest buildQueryRequest(
...
-      builder.setLimit(DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
-      requestContext.setOrderByExpressions(getRequestOrderByExpressions(request));
-    } else { // No Group By: Use Pinot's Order By, Limit and Offset
-      addSortLimitAndOffset(request, builder);
+    // 4. Add Order By
+    addSort(request, builder);
+
+    // 5. Set Limit, Offset
+    if (requestContext.hasGroupBy()) {
+      // providing both group by and order by doesn't provide right results in pinot
```
aaron-steinfeld, line 106: 3 concerns:

1. Does this break support for the "other" flag. E.g. if I'm saying give me 5 groups + other, do we need to over-query? We should make sure that's well tested. I don't think other makes sense when using offset, so not concerned about that - just the basic case of limit + other.
2. performance - forcing a query of limit+offset can easily run into exhaustion issues, moving a lot of data. At a minimum we should update our validation in this case that limit + offset is less than max limit, but I'd concern whether we really even want to support offset on a group by given the first concern. We could just throw if the two are used together.
3. In this function, we're applying offset in the db in QS in some cases, in others we're skipping. In the result post process function that this handler implements, we always skip (assume that it was not applied in QS). I don't have the code up outside the diff so not sure if that path is only hit in the group by case, but even if so, those types of assumptions make the code exceedingly error prone and complex to follow.

saxenakshitiz: 1. Validated the `other` flag, and it works fine with this change
2. Updated such that group by and offset results in unsupported operation exception
3. No longer an issue, given offset + group by are not supported

BTW I have done some testing on saas dev sandbox. How do we exhaustively test all possible scenarios?

aaron-steinfeld (commented):
Seems fine, could you update the PR title to match the change

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
aaron-steinfeld: All scenarios is hard :) I'd probably test a matrix of [different permutations of explorer params - time series, group by, group by w/ rest, grouped time series, grouped time series w/ rest, ordered group by etc.] x [different data sources - ES, QS]

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -95,21 +92,19 @@ QueryRequest buildQueryRequest(
...
-    if (requestContext
-        .hasGroupBy()) { // Will need to do the Ordering, Limit and Offset ourselves after we get
-      // the Group By Results
-      builder.setLimit(DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
-      requestContext.setOrderByExpressions(getRequestOrderByExpressions(request));
-    } else { // No Group By: Use Pinot's Order By, Limit and Offset
-      addSortLimitAndOffset(request, builder);
-    }
+    // 4. Add order by along with setting limit, offset
+    addSortLimitAndOffset(request, builder);
```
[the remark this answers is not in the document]
tim-mwangi: According to this comment https://github.com/hypertrace/gateway-service/blob/68ec6aabac770dcb76109d063c888cadb728df2e/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandlerWithSorting.java#L9 we do it in memory since at the time Pinot did not support Group By and Order By at the same time in a query. If this has changed, there's no need to keep doing it in memory.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
saxenakshitiz: Made all required code changes and fixed the tests as well. Please have a look.

aaron-steinfeld (dismissed):

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

PR #172, opened 2023-08-04
fix: fix to handle group by with offset

saxenakshitiz:
Group by query with offset doesn't retrieve right results from pinot. So here are the changes done -
1. In case offset is provided with group by query, we just add offset to limit itself when querying Pinot. We will just skip the results(from Pinot) till offset when returning final results to the caller.
2. The rest group doesn't make sense with pagination scenario(query with offset). Such query will result in UnsupportedOperationException
3. Ideally for pagination case, no group limit should be specified. In case it is specified, we will use limit as min(limit, group-limit) + offset to query pinot.

### DIFF
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
@@ -175,7 +175,14 @@ public class RequestHandler implements RequestHandlerWithSorting {
       // we will add offset to limit itself and then ignore results till offset in response
       limit += request.getOffset();
       // don't exceed default group by limit
-      limit = Math.min(limit, DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+      if (limit > DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT) {
+        LOG.error(
+            "Trying to query for rows more than the default limit {} : {}",
+            DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT,
+            request);
+        throw new UnsupportedOperationException(
+            "Trying to query for rows more than the default limit " + request);
+      }
       queryBuilder.setLimit(limit);
     } else {
       queryBuilder.setLimit(request.getLimit());
--- a/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
+++ b/gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
@@ -92,12 +92,15 @@ public class RequestHandler implements RequestHandlerWithSorting {
     builder.setFilter(
         constructQueryServiceFilter(request, requestContext, attributeMetadataProvider));
 
-    if (requestContext.hasGroupBy() && requestContext.getOffset() > 0) {
-      // providing both group by and offset doesn't provide right results in pinot
+    if (requestContext.hasGroupBy() && request.getIncludeRestGroup() && request.getOffset() > 0) {
+      // including rest group with offset is an invalid combination
       // throwing unsupported operation exception for this case
-      LOG.error("Query having group by along with offset is not supported : {}", request);
+      LOG.error(
+          "Query having group by with both offset and include rest is an invalid combination : {}",
+          request);
       throw new UnsupportedOperationException(
-          "Query with group by along with offset is not supported " + request);
+          "Query having group by with both offset and include rest is an invalid combination "
+              + request);
     }
 
     // 3. Add GroupBy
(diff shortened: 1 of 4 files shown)


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -160,13 +165,22 @@ private void addSortLimitAndOffset(
...
+    if (requestContext.hasGroupBy()) {
+      int limit = request.getLimit();
+      if (request.getGroupLimit() > 0) {
+        // in group by scenario, set limit to minimum of limit or group-limit
+        limit = Math.min(request.getLimit(), request.getGroupLimit());
+      }
+      // pinot doesn't handle offset with group by correctly
+      // we will add offset to limit itself and then ignore results till offset in response
+      limit += request.getOffset();
+      // don't exceed default group by limit
```
aaron-steinfeld, line 177: Can you throw in this case? Better an error than incorrect results

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -160,13 +165,22 @@ private void addSortLimitAndOffset(
...
+      limit = Math.min(limit, DEFAULT_QUERY_SERVICE_GROUP_BY_LIMIT);
+      queryBuilder.setLimit(limit);
     } else {
       queryBuilder.setLimit(request.getLimit());
+      queryBuilder.setOffset(request.getOffset());
```
aaron-steinfeld, line 182: So which path is this called in? If offset doesn't make sense for a time series or a group by/aggregation, does explorer support anything else?

And as a follow on to this, if some query paths do push down offset like this, how do we know in `sortAndPaginatePostProcess` below whether we should be doing it in memory or not?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
saxenakshitiz: dome

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
saxenakshitiz: > So which path is this called in? If offset doesn't make sense for a time series or a group by/aggregation, does explorer support anything else?
Not sure whether this else part will be accessed by explorer query. However such queries can be constructed and accessed via graphql. We should ensure that path works.
 
> And as a follow on to this, if some query paths do push down offset like this, how do we know in sortAndPaginatePostProcess below whether we should be doing it in memory or not? 
`sortAndPaginatePostProcess` was created to handle group by case only. 
Group by query ideally with no group limit and optional offset. This query can have aggregations as well. This is the case where we are handling offset in memory. We will not allow `rest group` to be used in this query.
I am assuming time series query will not have offset. It will just have limit, group-limit and rest-group.

aaron-steinfeld: > Not sure whether this else part will be accessed by explorer query. However such queries can be constructed and accessed via graphql. We should ensure that path works.

Correct, that's what I'm asking. Time series have their own handler (but I think it extends this one - what a mess!) so that shouldn't be an issue, but if there is any query for which this path is reachable (which I believe would be an explorer query without aggregations aka an event/span query), then that would mean the new code added on line 341 would apply the offset in memory which has already been pushed down. And if it's not reachable we shouldn't have it - either way, the two methods aren't consistent.

saxenakshitiz: Only in case of group by query with offset, we update the limit and then skip those results. So it will be consistent.

For time aggregations, there will always be a group by so above logic will work consistently in that case as well.

However offset in time aggregations query is certainly a misfit.

aaron-steinfeld: Pulled down the code and walked through the paths to it. Looks like timeseries has its own, entitiy doesn't use it. So for the remaining scenarios it appears correct, but this is so convoluted it's just asking for the next person to break it :(

- group by w/ rest + no offset (correct, offset zero avoids the in memory pagination)
- group by w/ rest  + offset (now throws)
- group by w/o rest + no offset (correct, offset zero avoids in memory pagination)
- group by w/o rest + offset (correct, paginates in memory)
- no group by (pushes down offset, if present)

aaron-steinfeld (approved):

PR #175, opened 2023-09-11
Fetch entity types from entity service

anujgoyal1:
## Description
Please include a summary of the change, motivation and context.

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:

### Documentation


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EntityTypesProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(EntityTypesProvider.class);
+  private static final int DEFAULT_CACHE_SIZE = 1024;
+  private static final int DEFAULT_EXPIRE_DURATION_MIN = 30; // 30 minutes
+
+  // TenantId to Entity types cache
+  private final LoadingCache<String, List<String>> tenantIdToEntityTypesCache;
```
skjindal93, line 24: Can we use request context as cache key, instead of tenant?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public EntityTypesProvider(EntityTypeServiceClient entityTypeServiceClient) {
+    CacheLoader<String, List<String>> cacheLoader =
+        new CacheLoader<>() {
+          @Override
+          public List<String> load(@NonNull String tenantId) {
+            return entityTypeServiceClient.getAllEntityTypes(tenantId).stream()
+                .map(EntityType::getName)
+                .collect(Collectors.toList());
```
skjindal93, line 33:
```suggestion
                .collect(Collectors.toUnmodifiableList());
```

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,55 @@
...
+    tenantIdToEntityTypesCache =
+        CacheBuilder.newBuilder()
+            .maximumSize(DEFAULT_CACHE_SIZE)
+            .expireAfterWrite(DEFAULT_EXPIRE_DURATION_MIN, TimeUnit.MINUTES)
+            .build(cacheLoader);
+  }
+
+  public List<String> getEntityTypes(String tenantId) {
+    if (StringUtils.isBlank(tenantId)) {
+      LOG.error("Tenant id not found while trying to fetch entity types");
```
aman-bansal, line 46: shouldnt we return or throw an error here? whats the behaviour in this case? do we have any default case?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
skjindal93, line 24: Can we cache it as a set, instead of a list?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -103,14 +109,18 @@ public ExploreResponse explore(RequestContext requestContext, ExploreRequest req
     }
   }
 
-  private boolean isContextAnEntityType(ExploreRequest request) {
-    return Arrays.stream(EntityType.values())
-        .anyMatch(entityType -> entityType.name().equalsIgnoreCase(request.getContext()));
+  private boolean isContextAnEntityType(ExploreRequest request, String tenantId) {
+    return Stream.concat(
+            entityTypesProvider.getEntityTypes(tenantId).stream(),
```
aman-bansal, line 114: wouldnt entity types provider will provide the complete list of types? do we need stream of enums?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -91,7 +96,8 @@ public ExploreResponse explore(RequestContext requestContext, ExploreRequest req
               newExploreRequestContext, request.getContext());
       exploreRequestValidator.validate(request, attributeMetadataMap);
 
-      IRequestHandler requestHandler = getRequestHandler(request, attributeMetadataMap);
+      IRequestHandler requestHandler =
+          getRequestHandler(request, attributeMetadataMap, requestContext.getTenantId());
```
aman-bansal, line 100: generally we should only send request context and can use contextual key as the key for cache.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
anujgoyal1: What's will be the advantage of that? 🤔

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
anujgoyal1: It should. @skjindal93 is it safe to remove the dependency on the enums?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
anujgoyal1: Since we are relying on the enum based too, I had not thrown error and just logged it, so old flow can still support the queries.
If we want to remove enum based entity types altogether, we can throw the error here

aman-bansal (approved):

skjindal93 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/util/EntityTypeServiceClient.java
```diff
@@ -0,0 +1,27 @@
...
+
+  public EntityTypeServiceClient(Channel channel) {
+    this.blockingStub =
+        EntityTypeServiceGrpc.newBlockingStub(channel)
+            .withCallCredentials(
+                RequestContextClientCallCredsProviderFactory.getClientCallCredsProvider().get());
+  }
+
+  public List<EntityType> getAllEntityTypes(RequestContext requestContext) {
+    return Lists.newArrayList(
```
aaron-steinfeld, line 23: Why are you intentionally converting this into a mutable list?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public Set<String> getEntityTypes(ContextualKey<Void> contextualKey) {
+    try {
+      return tenantIdToEntityTypesCache.get(contextualKey);
+    } catch (ExecutionException e) {
+      LOG.error(
+          String.format(
+              "Error retrieving entity types for tenant %s",
+              contextualKey.getContext().getTenantId()));
```
aaron-steinfeld, line 57: Log context rather than tenant

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+      throw new RuntimeException(
```
aaron-steinfeld, line 58: Throw grpc exceptions rather than plain runtime

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+          @Override
+          public Set<String> load(@NonNull ContextualKey<Void> contextualKey) {
+            Set<String> entityTypesSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+            entityTypeServiceClient
+                .getAllEntityTypes(contextualKey.getContext())
+                .forEach(entityType -> entityTypesSet.add(entityType.getName()));
+            entityTypeServiceV2Client
+                .getAllEntityTypes(contextualKey.getContext())
+                .forEach(entityType -> entityTypesSet.add(entityType.getName()));
+            return entityTypesSet;
```
aaron-steinfeld, line 39: immutable set for a provider please - also any reason for tree set here? Are we relying on them being ordered by name?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+import org.hypertrace.gateway.service.common.util.EntityTypeServiceV2Client;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EntityTypesProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(EntityTypesProvider.class);
+  private static final int DEFAULT_CACHE_SIZE = 1024;
+  private static final int DEFAULT_EXPIRE_DURATION_MIN = 30; // 30 minutes
+
+  // TenantId to Entity types cache
```
aaron-steinfeld, line 22: the tenant ID part should be opaque, that's why we're using a context key. Suggest just calling this an entityTypeCache

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+    tenantIdToEntityTypesCache =
+        CacheBuilder.newBuilder()
+            .maximumSize(DEFAULT_CACHE_SIZE)
+            .expireAfterWrite(DEFAULT_EXPIRE_DURATION_MIN, TimeUnit.MINUTES)
+            .build(cacheLoader);
+  }
+
+  public Set<String> getEntityTypes(ContextualKey<Void> contextualKey) {
```
aaron-steinfeld, line 50: Accept a context rather than a contextual key. By asking for a key, you're forcing the caller to decide how the cache should be determined which is an implementation detail to this provider (as is the fact that it caches at all).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/EntityTypesProvider.java
```diff
@@ -0,0 +1,64 @@
...
+  // TenantId to Entity types cache
+  private final LoadingCache<ContextualKey<Void>, Set<String>> tenantIdToEntityTypesCache;
+
+  public EntityTypesProvider(
+      EntityTypeServiceClient entityTypeServiceClient,
+      EntityTypeServiceV2Client entityTypeServiceV2Client) {
+    CacheLoader<ContextualKey<Void>, Set<String>> cacheLoader =
+        new CacheLoader<>() {
+          @Override
+          public Set<String> load(@NonNull ContextualKey<Void> contextualKey) {
```
aaron-steinfeld, line 31: Try to avoid inlining methods like this. Just `CacheLoader.from(this::fetch)` for example.

Also consider making this async - easily can avoid the extra remote call on fetch path.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -103,14 +108,21 @@ public ExploreResponse explore(RequestContext requestContext, ExploreRequest req
     }
   }
 
-  private boolean isContextAnEntityType(ExploreRequest request) {
+  private boolean isContextAnEntityType(ExploreRequest request, RequestContext requestContext) {
+    if (entityTypesProvider
```
aaron-steinfeld, line 112: This should wholly replace the prior impl which remains on lines 117-119. If the enum has a value that's not registered, it's not a real entity type and should not be treated as such.

PR #176, opened 2023-09-14
fix(explore): skip child filters for default filter

skjindal93:


anujgoyal1 (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -93,12 +93,18 @@ private void addSelections(
 
   private Filter.Builder buildFilter(
       ExploreRequest exploreRequest, List<String> entityIdAttributeIds, Set<String> entityIds) {
+    if (org.hypertrace.gateway.service.v1.common.Filter.getDefaultInstance()
+        .equals(exploreRequest.getFilter())) {
```
aaron-steinfeld, line 97: Can we rely on presence check? or is there an upstream bug? Either way, not sure I'm following why entity IDs are ignored if there's no filter? Shouldn't we just not be and'ing and using only the in filter?

PR #177, opened 2023-09-15
fix: fix time aggregations query having order by as well as order by based on interval

saxenakshitiz:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -107,9 +111,24 @@ private ExploreRequest buildGroupByRequest(ExploreRequest originalRequest) {
         .getTimeAggregationList()
         .forEach(timeAggregation -> requestBuilder.addSelection(timeAggregation.getAggregation()));
 
+    // remove interval based order by expression(if present) from order by expressions in the
+    // groupBy
+    // request being generated
```
skjindal93, line 116:
```suggestion
    // remove interval based order by expression(if present) from order by expressions in the
    // groupBy request being generated
```

skjindal93: Can we add the reason to remove the interval based order by in case of group by, as a comment?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -107,9 +111,24 @@ private ExploreRequest buildGroupByRequest(ExploreRequest originalRequest) {
...
+    List<OrderByExpression> orderByExpressionList =
```
skjindal93, line 117: Does it mean that it won't be ordered by interval start time anyways, even if it is present as order by?

aaron-steinfeld: Yeah this seems off. I would think we'd use a default order by if none provided (ascending time) but if one is requested, use that only. If, for backwards compatibility we need to shift that approach slightly we can (e.g. always apply some time order), but any order bys given by caller should have priority.

saxenakshitiz: The time aggregations with group by query is split in two part. First query gets the relevant groups. In this query interval based grouping is not required. This is the current change.
Once groups are queried, then following does time aggregations. In this query, we have both group by and order by based on interval defined.
This is covered in test updated. Please check -
request1(query to get groups which should not have interval based groupby or orderby) - https://github.com/hypertrace/gateway-service/blob/525fc0d1ee04047185d3d5d55ff776710071acae/gateway-service-impl/src/test/resources/query-service-requests-and-responses/explore/grouped-time-aggregation-with-group-limit.json#L3
request2 (query to get final response  which should have interval based groupby or orderby) -  https://github.com/hypertrace/gateway-service/blob/525fc0d1ee04047185d3d5d55ff776710071acae/gateway-service-impl/src/test/resources/query-service-requests-and-responses/explore/grouped-time-aggregation-with-group-limit.json#L108

aaron-steinfeld: Got it. So we have no way of specifying an ordering on how to select the groups to use right?

saxenakshitiz: Q1 is correct
Q2 will be service name, aggregate-function-for-call-count `group by` service name, date-time-convert-function `order by` date-time-convert-function, aggregate-function-for-call-count
date-time-convert-function - function that will round time to minute
aggregate-function-for-call-count - aggregate function defined on call count

aaron-steinfeld (approved):
LGTM other than dep check issue

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
saxenakshitiz: Added

aaron-steinfeld (approved):

PR #179, opened 2023-09-19
chore: added error log

saxenakshitiz:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -102,6 +109,21 @@ public ExploreResponse explore(RequestContext requestContext, ExploreRequest req
           requestHandler.handleRequest(newExploreRequestContext, request);
 
       return responseBuilder.build();
+    } catch (Exception ex) {
+      boolean acquireSucceeded = rateLimiter.tryAcquire();
+      if (acquireSucceeded || log.isDebugEnabled()) {
```
aaron-steinfeld, line 114: In general, try to avoid embedding error logs and rethrowing. Catch and craft a new error if needed, but always error log in one place - the service entry. Otherwise you're just going to duplicate, because we need to log any thrown errors there anyway.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/ExploreService.java
```diff
@@ -102,6 +109,21 @@ public ExploreResponse explore(RequestContext requestContext, ExploreRequest req
           requestHandler.handleRequest(newExploreRequestContext, request);
 
       return responseBuilder.build();
+    } catch (Exception ex) {
+      boolean acquireSucceeded = rateLimiter.tryAcquire();
```
aaron-steinfeld, line 113: Check debug before trying to acquire rate limiter (which is an expensive operation).

aaron-steinfeld (approved):

PR #183, opened 2023-11-22
chore | add list interactions filters in entity service

aman-bansal:


gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/entities.proto
```diff
@@ -38,8 +38,10 @@ message EntitiesRequest {
   repeated org.hypertrace.gateway.service.v1.common.Expression selection = 5;
   repeated org.hypertrace.gateway.service.v1.common.TimeAggregation time_aggregation = 6;
   // Interaction requests
-  InteractionsRequest incoming_interactions = 7;
-  InteractionsRequest outgoing_interactions = 8;
+  InteractionsRequest incoming_interactions = 7 [deprecated = true];
+  InteractionsRequest outgoing_interactions = 8 [deprecated = true];
+  repeated InteractionsRequest incoming_interaction_requests = 9;
```
aaron-steinfeld, line 43: What's the actual goal here - to be able to put different selections for different filters? And if so, are we just dumping them all together in the response for the client to sort out? This seems questionable, particularly without a response change

aman-bansal: no the goal is to have different filters for different interaction requests. This is needed for new application flow changes. Currently, GraphQL combines everything into one interaction request which results in the wrong filter combination. For example, while querying outgoing interactions with the service and backend, filters get messed up and lead to wrong results.

aaron-steinfeld: If we're not changing the selections or limits, doesn't the current filter tree already support that? At least I'm hearing something like 

`OR( AND( type = service, other service filters), AND(type = backend, other backend filters))`

PR #186, opened 2023-12-06
chore: convert 'null' string array value to list with 'null' value

saxenakshitiz:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/StringToAttributeKindConverter.java
```diff
@@ -92,6 +94,12 @@ private List<String> convertToArray(String jsonString) {
       return List.of();
     }
 
+    // handle special case when "null" string is returned as string array value(default value
```
aaron-steinfeld, line 97: Discussed offline - concerned about this behavior change, as every query that was previously returning an empty array (e.g. give me the labels for this API) will now start containing nulls. Existing behavior seemed more accurate, we just need to make sure we appropriately handle group by case and either return no groups (short circuit) or translate it back if we think we should look for explicitly empty groups as part of a group by query. I'm not sure what the actual correct expectation should be there.

aaron-steinfeld (approved):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -228,7 +230,14 @@ private void buildInClauseFilterValue(Value value, Value.Builder valueBuilder) {
         valueBuilder.addStringArray(value.getString());
         break;
       case STRING_ARRAY:
-        valueBuilder.addAllStringArray(value.getStringArrayList());
+        // Handle special case when "null" string is returned as string array value(default value
+        // scenario). This will be converted to empty list in value converter. In such a case use
+        // list with "null" value
+        if (value.getStringArrayList().isEmpty()) {
```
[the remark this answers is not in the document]
kotharironak: > > so one behavior discrepancy this would lead to to - let's say we have 10 rows, and 9 of them have no values for a multivalued col X and 1 of them does. Both before and after this change, if we grouped by X we'd get a single returned grouped with a count of 1.
> 
> From the Query Service (QS) point of view, it will return two groups. Let's assume that 1 row which is not empty has value - ["hello"]. The output will be as below. Returned groups:
> 
> ```
> {
>   {"null", 9},
>   {"hello", 1}
> }
> ```
> 
> And if the none empty row of X has more than one value, say - ["hello", "world"], QS will return 3 groups:
> 
> ```
> {
>   {"null", 9},
>   {"hello", 1},
>   {"world", 1}
> }
> ```
> 
> Moving on to Pinot, if the column (or multi-value column) has no value, it will return the default value. The default value for a String column is `null`. For a multi-value column, if not preset, it uses the default of a single-value column of the same type.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
```diff
@@ -304,4 +306,33 @@ private Map<String, AttributeMetadata> remapAttributeMetadataByResultName(
             .collect(Collectors.toUnmodifiableList()),
         attributeMetadataByIdMap);
   }
+
+  private void updateExploreResponse(ExploreResponse.Builder timeAggregationsResponse) {
```
kotharironak, line 310: Isn't this changing the existing output behaviour? So, we never returned the `null` group?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -224,10 +226,28 @@ private Set<String> getExcludedValues(
...
-        .map(Value::getString)
+        .flatMap(this::getStringValues)
         .collect(ImmutableSet.toImmutableSet());
   }
 
+  private Stream<String> getStringValues(Value value) {
+    switch (value.getValueType()) {
+      case STRING:
+        return Stream.of(value.getString());
+      case STRING_ARRAY:
```
kotharironak, line 237: Can you change the comment?
```
// The downstream QS service returns the default value "null" for a string array when it is empty.
// GW returns an empty list as output. This transformation occurs in the value converter.
// To handle this scenario when querying downstream with an empty list, set its default value to "null."
```

gateway-service-impl/src/test/resources/expected-test-responses/explore/grouped-time-aggregation-with-group-limit-with-null-value-group.json
```diff
@@ -0,0 +1,45 @@
+{
+  "row": [{
```
kotharironak, line 2: Whats the query look like for this for gateway?

kotharironak: Oh! okey, it will be an internal query. Or would be IN query from GraphQL to gateway too. That is currently not expressible, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
saxenakshitiz: Updated

kotharironak (commented):
LGTM.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TimeAggregationsWithGroupByRequestHandler.java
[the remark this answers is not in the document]
aaron-steinfeld: The described behavior makes sense, the first issue to resolve however is how GW should behave because it's not consistent. If we do a group by with time series, it won't return that null (or empty list); if we do it without, it returns an empty list as one value.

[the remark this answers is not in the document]
aaron-steinfeld: Agreed that both should return empty list

gateway-service-impl/src/test/resources/attributes/attributes.json
```diff
@@ -946,9 +946,9 @@
     ]
   },
   {
-    "fqn": "Service.userRoles",
+    "fqn": "Service.labels",
     "valueKind": "TYPE_STRING_ARRAY",
-    "key": "userRoles",
+    "key": "labels",
     "displayName": "Service User Roles",
```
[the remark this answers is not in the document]
saxenakshitiz: Done

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/TheRestGroupRequestHandler.java
```diff
@@ -224,10 +227,21 @@ private Set<String> getExcludedValues(
...
```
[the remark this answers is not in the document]
saxenakshitiz: That is correct. NOT_IN group was empty

gateway-service-impl/src/test/resources/expected-test-responses/explore/grouped-time-aggregation-with-group-limit-with-null-value-group.json
saxenakshitiz: We have all queries in the test
1. GraphQL query to GW
2. GW query to QS and its response
3. GW response to GraphQL

aaron-steinfeld (approved):

PR #193, opened 2024-02-21
perf(entities): optimise AND filter query

skjindal93:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -1,12 +1,19 @@
 package org.hypertrace.gateway.service.entity.query;
```
skjindal93, line 1: logic class

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -305,6 +316,115 @@ QueryNode buildFilterTree(EntitiesRequest entitiesRequest, Filter filter) {
     }
   }
 
+  QueryNode buildAndFilterTree(EntitiesRequest entitiesRequest) {
```
skjindal93, line 319: The most important logic

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -118,6 +119,19 @@ protected static EntityResponse union(List<EntityResponse> entityResponses) {
 
   @Override
   public EntityResponse visit(DataFetcherNode dataFetcherNode) {
+    QueryNode childNode = dataFetcherNode.getChildNode();
```
skjindal93, line 122: The most important logic

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -147,13 +155,19 @@ public QueryNode build() {
...
      * into one {@link DataFetcherNode}, instead of having multiple {@link DataFetcherNode}s for
      * each filter
+     *
+     * <p>It is not needed for AND filter, since the filter tree is already optimised with a single
+     * data fetcher node for each source
      */
-    QueryNode optimizedFilterTree = filterTree.acceptVisitor(new FilterOptimizingVisitor());
-    if (LOG.isDebugEnabled()) {
-      LOG.debug("Optimized Filter Tree:{}", optimizedFilterTree.acceptVisitor(new PrintVisitor()));
+    boolean isAndFilter = executionContext.getExpressionContext().isAndFilter();
```
avinashkolluru, line 162: Didnt understand this. The comment says optimization visitor is not needed for AND filter but the condition is checking for isAndFilter. I'm confused.

skjindal93: Thanks for catching it. I have fixed it. It should have been `!isAndFilter`

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
```diff
@@ -19,13 +19,26 @@ public class DataFetcherNode implements QueryNode {
...
+    this.childNode = new NoOpNode();
+
+    // total would be computed in memory
+    this.canFetchTotal = false;
+  }
+
+  public DataFetcherNode(String source, Filter filter, QueryNode childNode) {
+    this.source = source;
+    this.filter = filter;
+    this.childNode = childNode;
```
kotharironak, line 38: Is there any restriction or limitation on the type of data a childNode fetch node can have?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -147,13 +155,19 @@ public QueryNode build() {
      * {@link FilterOptimizingVisitor} is needed to merge filters corresponding to the same source
      * into one {@link DataFetcherNode}, instead of having multiple {@link DataFetcherNode}s for
      * each filter
+     *
+     * <p>It is not needed for AND filter, since the filter tree is already optimised with a single
```
kotharironak, line 159: Is the same true for `OR` filter too?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -305,6 +316,104 @@ QueryNode buildFilterTree(EntitiesRequest entitiesRequest, Filter filter) {
     }
   }
 
+  QueryNode buildAndFilterTree(EntitiesRequest entitiesRequest) {
+    // If the filter by and order by are from QS, pagination can be pushed down to QS
+    // Since the filter and order by are from QS, there won't be any filter on other
```
kotharironak, line 321: Isn't the single source fetch is already handled at the beginning of the function?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -305,6 +316,104 @@ QueryNode buildFilterTree(EntitiesRequest entitiesRequest, Filter filter) {
...
+          entitiesRequest.getFetchTotal(),
+          qsNode);
+    } else {
+      // if there are multiple filter sources
+      List<QueryNode> dataFetcherNodes = new ArrayList<>();
+      for (Map.Entry<AttributeSource, Filter> entry : sourceToAndFilterMap.entrySet()) {
+        AttributeSource source = entry.getKey();
+        Filter andFilter = entry.getValue();
+
+        dataFetcherNodes.add(new DataFetcherNode(source.name(), andFilter, qsNode));
```
kotharironak, line 379: This make sure that first we get the data from QS as its child node, correct?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -118,6 +119,19 @@ protected static EntityResponse union(List<EntityResponse> entityResponses) {
...
+
+    EntityResponse childNodeResponse = dataFetcherNode.getChildNode().acceptVisitor(this);
+    EntityFetcherResponse childEntityFetcherResponse = null;
+    Filter childFilter;
+    if (childNode instanceof NoOpNode) {
+      childFilter = Filter.getDefaultInstance();
+    } else {
+      // Construct the filter from the child nodes result
+      childEntityFetcherResponse = childNodeResponse.getEntityFetcherResponse();
```
kotharironak, line 131: Are we restricting the max number of entities in the sub-sequent filter?

avinashkolluru:
@skjindal93 other change look fine to me. But, the build has been failing. Please check and fix.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/config/GatewayServiceConfig.java
```diff
@@ -0,0 +1,39 @@
+package org.hypertrace.gateway.service.common.config;
+
+import com.typesafe.config.Config;
+import org.hypertrace.gateway.service.entity.config.EntityIdColumnsConfig;
+import org.hypertrace.gateway.service.entity.config.LogConfig;
+
+public class GatewayServiceConfig {
+  private static final String ENTITY_AND_FILTER_ENABLED_CONFIG_KEY = "filter.entity.and.enabled";
```
kotharironak, line 8: `entity.andFilter.enabled` looks more readable.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/DataFetcherNode.java
```diff
@@ -19,13 +19,26 @@ public class DataFetcherNode implements QueryNode {
...
     this.source = source;
     this.filter = filter;
-    this.canFetchTotal = false; // total would be computed in memory
+    this.childNode = new NoOpNode();
+
+    // total would be computed in memory
+    this.canFetchTotal = false;
+  }
+
+  public DataFetcherNode(String source, Filter filter, QueryNode childNode) {
```
kotharironak, line 35: As mentioned prior, Should we be restrictive that DataFetcherNode can have only another DataFetcherNode as child or NoOp node?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -305,6 +318,106 @@ QueryNode buildFilterTree(EntitiesRequest entitiesRequest, Filter filter) {
...
+    // sources
+    if (sourceSetsIfFilterAndOrderByAreFromSameSourceSets.contains(QS.name())) {
+      executionContext.setSortAndPaginationNodeAdded(true);
+      return createQsDataFetcherNodeWithLimitAndOffset(entitiesRequest);
+    }
+
+    Map<AttributeSource, Filter> sourceToAndFilterMap =
+        new EnumMap<>(buildSourceToAndFilterMap(entitiesRequest.getFilter()));
+
+    // qs node as the pivot node to fetch time range data
```
kotharironak, line 335: So, there won't be any query w/o time range filter, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -128,6 +140,15 @@ public EntityResponse visit(DataFetcherNode dataFetcherNode) {
             entitiesRequest.getEntityType(),
             executionContext.getTimestampAttributeId());
 
+    Filter entitiesRequestFilter =
+        childFilter != null
```
kotharironak, line 144: When result is empty - it seems that the method `constructFilterFromChildNodesResult` returns `Filter.getDefaultInstance()`
Is the defaultInstance equals `null`?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -160,17 +181,28 @@ public EntityResponse visit(DataFetcherNode dataFetcherNode) {
...
+              : entityFetcherResponse;
+      return new EntityResponse(response, totalFuture.join());
     } else {
       // if the data fetcher node is not paginating, the total number of entities is equal to number
       // of records fetched
-      EntityFetcherResponse response = entityFetcher.getEntities(context, request);
+      EntityFetcherResponse entityFetcherResponse = entityFetcher.getEntities(context, request);
+      response =
+          childEntityFetcherResponse != null
+              ? intersectEntities(List.of(childEntityFetcherResponse, entityFetcherResponse))
```
kotharironak, line 204: Curious Q: Do we still need to intersect when `childEntityFetcherResponse` is `not null`? In that case, wouldn't the child filter have already applied in subsequent query?

gateway-service/src/main/resources/configs/common/application.conf
```diff
@@ -69,6 +69,10 @@ entity.idcolumn.config = [
   },
 ]
 
+filter.entity = {
```
kotharironak, line 72: entity.andFilter.enabled?

kotharironak (commented):
Overall looks fine.

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
skjindal93: > Is the defaultInstance equals null?

No. It's just an empty filter instance

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
skjindal93: `intersectEntities` also takes care of merging the selection attributes fetched from various sources

If we fetch attribute A from QS, and attribute B from EDS for the same entity, `intersectEntities` would merge the responses, and have both A and B added as an attribute on the entity response

gateway-service/src/main/resources/configs/common/application.conf
skjindal93: I usually follow a top-down hierarchial approach while building configs, so that the configs are re-usable at a parent level

In this case, `filter` being the parent can be extended to other configs, and so on

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
skjindal93: right

gateway-service/src/main/resources/configs/common/application.conf
kotharironak: While reading the code, all variable names are isEntityAndFilterEnabled, but the corresponding configuration name differs in the application.conf file. This makes it a bit difficult to easily co-relate.

kotharironak (approved):

PR #196, opened 2024-03-07
fix(explore): return early, if no entities

skjindal93:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -118,7 +123,13 @@ QueryRequest buildQueryRequest(
               .orElse(request.getFilter());
     }
 
+    if (entityIds.isEmpty()) {
```
aaron-steinfeld, line 126: entityIds starts empty and only gets filled if the condition on 119 is met - so that means that this would always short-circuit if we have a metric source, right?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
```diff
@@ -493,6 +493,10 @@ public static Filter addTimeSpaceAndIdFiltersAndConvertToQueryFilter(
       List<String> entityIdAttributes,
       org.hypertrace.gateway.service.v1.common.Filter providedFilter) {
 
+    if (entityIds.isEmpty()) {
```
aaron-steinfeld, line 496: Do we still need this? Can we avoid the default instance filter?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
skjindal93: good catch. moved the empty entity ids check inside the attribute source condition

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/common/converters/QueryAndGatewayDtoConverter.java
skjindal93: removed it

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -113,12 +118,18 @@ QueryRequest buildQueryRequest(
         attributeMetadataProvider.getAttributesMetadata(requestContext, request.getContext());
     if (hasOnlyAttributeSource(request.getFilter(), AttributeSource.EDS, attributeMetadataMap)) {
       entityIds = getEntityIdsToFilterFromSourceEDS(requestContext, request, attributeMetadataMap);
+
+      if (entityIds.isEmpty()) {
```
aaron-steinfeld, line 122: nit - worth a comment explaining why we shouldn't build a query request in this case

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -88,16 +88,21 @@ public RequestHandler(
...
       ExploreRequestContext requestContext, ExploreRequest request) {
-    QueryRequest queryRequest =
+    Optional<QueryRequest> maybeQueryRequest =
         buildQueryRequest(requestContext, request, attributeMetadataProvider);
 
+    if (maybeQueryRequest.isEmpty()) {
+      return ExploreResponse.newBuilder();
+    }
+
+    QueryRequest queryRequest = maybeQueryRequest.get();
```
aaron-steinfeld, line 98: nit - prefer `.orElseThrow` over `.get` (or move code around to avoid the direct access)

aaron-steinfeld (approved):
minor comments

PR #197, opened 2024-03-08
fix | fix the entity ids filter for explore if empty

aman-bansal:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -532,6 +538,73 @@ private Map<String, AttributeMetadata> remapAttributeMetadataByResultName(
         attributeMetadataByIdMap);
   }
 
+  private org.hypertrace.gateway.service.v1.common.Filter createEntityIdAttributeFilter(
+      List<String> entityIdAttributes,
+      org.hypertrace.gateway.service.v1.common.Operator operator,
+      List<String> entityIds) {
+    if (entityIdAttributes.size() != 1) {
+      throw new RuntimeException("Cannot have more than one id attribute for an entity");
```
aaron-steinfeld, line 546: nit - throw grpc status exception please

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -532,6 +538,73 @@ private Map<String, AttributeMetadata> remapAttributeMetadataByResultName(
...
+    }
+
+    if (entityIds.isEmpty()) {
+      // Having empty entity ids is valid filter because this means that
```
aaron-steinfeld, line 550: It's fine as a quick fix for the issue, but this is not a great long term solution to fire a query we know will return empty. Can we leave a TODO to come up with a better solution here?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -532,6 +538,73 @@ private Map<String, AttributeMetadata> remapAttributeMetadataByResultName(
...
+      // information as per the selections they have sent. So we are converting
+      // empty entity Id filter into a false filter (id != null && id == null)
+      // so that results are always empty but must contain the valid selections.
+      return org.hypertrace.gateway.service.v1.common.Filter.newBuilder()
+          .setOperator(Operator.AND)
+          .addChildFilter(
+              org.hypertrace.gateway.service.v1.common.Filter.newBuilder()
+                  .setLhs(buildStringExpression(entityIdAttributes.get(0)))
+                  .setOperator(Operator.NEQ)
+                  .setRhs(buildStringExpression(null))
```
aaron-steinfeld, line 565: we have a null literal expression for QS I thought? I guess it doesn't really matter, unless pinot optimizes for it

aman-bansal: We have it but at the query service layer. This is gateway service contract. I haven't tested it yet. Was waiting on whether this change is making sense or not

aman-bansal: Verified parsing of string null values is being handled by query service

aaron-steinfeld (approved):

PR #199, opened 2024-05-14
chore: configure hpa behavior

ravisingal:


helm/values.yaml
```diff
@@ -97,3 +97,19 @@ hpa:
   minReplicas: 1
   maxReplicas: 5
   targetCPUUtilizationPercentage: 80
+  behavior:
+    enabled: true
+    scaleUp:
+      policy:
+        type: Pods
+        value: 2
```
satish-mittal, line 105: shouldn't the max/min values in L105 and L112 align with existing `minReplicas` and `maxReplicas`?

satish-mittal, line 105: @ravisingal In simple words, this scaleUp policy will add max 2 pods within any 60 sec (until it reaches maxReplicas) and scaleDown policy will remove min 1 pod within any 600 sec (until it reaches minReplicas) after looking at previous window of 600 seconds, right?

satish-mittal (approved):

tim-mwangi (approved):

PR #201, opened 2024-05-21
chore: use filter in entities query for selections as well

saxenakshitiz:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/ExecutionTreeBuilder.java
```diff
@@ -330,7 +318,7 @@ QueryNode buildAndFilterTree(EntitiesRequest entitiesRequest) {
     }
 
     Map<AttributeSource, Filter> sourceToAndFilterMap =
-        new HashMap<>(buildSourceToAndFilterMap(entitiesRequest.getFilter()));
+        Map.copyOf(context.getExpressionContext().getSourceToFilterMap());
```
skjindal93, line 321: returns an unmodifiable map. The map is being modified on L328

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -325,7 +327,7 @@ public EntityResponse visit(SelectionNode selectionNode) {
                                   .getExpressionContext()
                                   .getSourceToTimeAggregationMap()
                                   .get(source))
-                          .setFilter(filter)
+                          .setFilter(addSourceFilters(executionContext, source, filter))
```
skjindal93, line 330: What about total? Should we send the same filter for the total request?

saxenakshitiz: total is fetched as part of data fetcher node, so all filters will be used

PR #202, opened 2024-06-06
chore | adding the support pagination in explore

aman-bansal:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -107,9 +111,19 @@ public ExploreResponse.Builder handleRequest(
...
     Iterator<ResultSetChunk> resultSetChunkIterator = executeQuery(requestContext, queryRequest);
+    ExploreResponse.Builder responseBuilder =
+        handleQueryServiceResponse(
+            request,
+            requestContext,
+            resultSetChunkIterator,
+            requestContext,
+            attributeMetadataProvider);
+
+    if (request.getFetchTotal()) {
```
suresh-prakash, line 122: Doing this in sequence is going to double-up the latency.
I guess, it's fine as long as the total latency is acceptable.
If required, can explore parallelising this.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
```diff
@@ -47,6 +48,7 @@ message EntityOption {
 
 message ExploreResponse {
   repeated org.hypertrace.gateway.service.v1.common.Row row = 2;
+  int32 total = 3;
```
suresh-prakash, line 51: Don't think, we would ever cross the `int32` limits, but, it's better to have it `int64` just in case we have an entity that has so many rows (perhaps in the future).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -227,6 +241,62 @@ private List<String> getEntityIdsToFilterFromSourceEDS(
...
+      }
+
+      if (!chunk.hasResultSetMetadata()
+          || chunk.getResultSetMetadata().getColumnMetadataList().size() != 1
+          || !TOTAL_ALIAS_NAME.equals(
+              chunk.getResultSetMetadata().getColumnMetadata(0).getColumnName())) {
+        LOG.warn("Chunk doesn't have result metadata so couldn't process the response.");
+        break;
+      }
+      return chunk.getRow(0).getColumn(0).getInt();
```
suresh-prakash, line 262: Don't think, we would ever cross the `int` limits, but, it's better to have it `getLong` just in case we have an entity that has so many rows (perhaps in the future).

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
```diff
@@ -227,6 +241,62 @@ private List<String> getEntityIdsToFilterFromSourceEDS(
...
+    }
+    return 0;
+  }
+
+  private QueryRequest buildTotalQueryRequest(ExploreRequest request, QueryRequest queryRequest) {
+    QueryRequest.Builder builder = QueryRequest.newBuilder();
+    List<org.hypertrace.gateway.service.v1.common.Expression> groupBys =
+        ExpressionReader.getAttributeExpressions(request.getGroupByList());
+    // add all group-by columns as distinct count to get total
```
suresh-prakash, line 271: Perhaps, I'm missing something here. Simply removing the sorting and pagination and getting the total count is sufficient, right?

Any reason why we want to pick the `group-by` columns? In a query, say `SELECT a, b FROM x WHERE ... GROUP BY a, b`, how does DISTINCT_COUNT(a) + DISTINCT_COUNT(b) give the total number of rows that query would return? Wouldn't the rows having both a and b be double-counted?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityRequestHandler.java
```diff
@@ -76,21 +73,15 @@ public ExploreResponse.Builder handleRequest(
 
     builder.addAllRow(
         entityServiceEntityFetcher.getResults(requestContext, exploreRequest, entityIds));
-
-    // If there's a Group By in the request, we need to do the sorting and pagination ourselves.
```
suresh-prakash, line 80: Any reason why we have removed this?

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -155,6 +184,43 @@ private Map<String, AttributeMetadata> remapAttributeMetadataByResultName(
...
+            .collect(Collectors.toUnmodifiableList());
+    builder.addSelection(
+        Expression.newBuilder()
+            .setFunction(
+                Function.newBuilder()
+                    .setFunctionName(FunctionType.DISTINCTCOUNT.name())
+                    .addAllArguments(groupBysExpression)
+                    .setAlias(TOTAL_ALIAS_NAME)
+                    .build())
+            .build());
```
suresh-prakash, line 220: Same here.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
```diff
@@ -33,6 +33,7 @@ message ExploreRequest {
 
   string space_id = 22;
   optional ContextOption context_option = 23;
+  bool fetch_total = 24;
```
kotharironak, line 36: Why do we need this?

The total count will be retrieved using a COUNT(*) query with the same set of filters. The UI can execute this query in parallel if we want to display the total count in the UI.

Internally, we will implement it the similar way by executing two parallel queries.  But, it will be executed for each pagination request too.

We have overloaded the traces/spans API in the past and observed that for every pagination, it runs this query. So, if the UI fires the total query separately, it will be executed only once.

aman-bansal: This is needed to get the count of groupby. We won't be calculating again and again. UI will not fetch for total all the time, if total is not required. It will make separate query for total

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/RequestHandler.java
aman-bansal: Yeah lets try with sequential for now. I am planning for the UI to query total separately then the data call

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityRequestHandler.java
aman-bansal: I checked all the scenarios i dont think this is needed.

gateway-service-api/src/main/proto/org/hypertrace/gateway/service/v1/explore.proto
aman-bansal: actually found the way fixing this PR

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityServiceEntityFetcher.java
```diff
@@ -168,11 +167,24 @@ private EntityQueryRequest buildRequest(
             .setFilter(buildFilter(exploreRequest, entityIdAttributeIds, entityIds));
 
     addGroupBys(exploreRequest, builder);
+    addSortBy(exploreRequest, builder);
     addSelections(requestContext, exploreRequest, builder);
-    builder.setLimit(DEFAULT_ENTITY_REQUEST_LIMIT);
+    builder.setLimit(exploreRequest.getLimit());
```
saxenakshitiz, line 172: Minor nit: In general order followed while forming request is to add selections, followed by groups, then order by and then limit and offset.

saxenakshitiz (commented):
LGTM. @aman-bansal you mentioned that you will try out these changes manually first, and then we can merge this PR. Please let me know if that testing is completed.

saxenakshitiz (dismissed):

gateway-service-impl/src/main/java/org/hypertrace/gateway/service/explore/entity/EntityRequestHandler.java
```diff
@@ -67,7 +59,8 @@ public ExploreResponse.Builder handleRequest(
     ExploreResponse.Builder builder = ExploreResponse.newBuilder();
     Set<String> entityIds = new HashSet<>();
     Optional<EntityOption> maybeEntityOption = getEntityOption(exploreRequest);
-    if (requestOnLiveEntities(maybeEntityOption)) {
+    boolean requestOnLiveEntities = requestOnLiveEntities(maybeEntityOption);
```
skjindal93, line 62:
```suggestion
    boolean canRequestLiveEntities = canRequestLiveEntities(maybeEntityOption);
```

skjindal93 (approved):

PR #204, opened 2024-06-23
chore | adding the support of joining nodes for eds source query

aman-bansal:


gateway-service-impl/src/main/java/org/hypertrace/gateway/service/entity/query/visitor/ExecutionVisitor.java
```diff
@@ -209,6 +221,91 @@ public EntityResponse visit(DataFetcherNode dataFetcherNode) {
...
+                          // sourceToOrderByMetricExpressionMap, sourceToOrderByTimeAggregationMap
+                          .clearOrderBy()
+                          .clearLimit()
+                          .clearOffset()
+                          .addAllSelection(
+                              executionContext
+                                  .getExpressionContext()
+                                  .getSourceToSelectionExpressionMap()
+                                  .get(source))
+                          .setFilter(addSourceFilters(executionContext, source))
```
aman-bansal, line 267: should we add id node filter if say ids lenght is less than 10K or something?

